PR #6, opened 2020-10-18
Update build and env files for better local dev set up

stonefruit:


Dockerfile
```diff
@@ -20,6 +20,7 @@ WORKDIR /app/backend
 
 COPY backend/package.json .
 COPY backend/package-lock.json .
+COPY backend/config/ config/
```
stonefruit, line 23: Not sure if this is correct. Build was complaining of this missing, but if the image was building previously, then it might be a non-issue? I believe it is not being used since we are using database strings for sequelize.

stonefruit:
Verified auto migrations work with postgres.

Was not able to verify e2e due to lack of credentials.

Dockerfile
glencbz: Could you attach the exact message? That's a kinda weird because there's no config directory there.

README.md
```diff
@@ -1,11 +1,33 @@
...
+
+Backend runs on port `4000` while Frontend runs on port `3000` and can be run using the instructions below.
+
+### Server
+
+#### npm script
+
+```bash
+cd backend
+npm run start-local
```
glencbz, line 27: We have the `start:dev` command, was there another motivation behind this?

backend/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "tsc",
     "start": "npm run migration:migrate && node --enable-source-maps dist/index.js",
+    "start-local": "npm run migration:migrate && rm -rf ./dist && tsc && node --enable-source-maps dist/index.js",
```
glencbz, line 9: We have the `start:dev` command, was there another motivation behind this?

README.md
```diff
@@ -16,3 +38,50 @@ docker build . -t call-home
...
+docker-compose up
+```
+
+To use postgres as your development database:
+
+```bash
+# backend/.env
+...
+...
+DATABASE_URL=postgresql://postgres:password@db:5432/callhome
```
glencbz, line 60: This should be in the docker-compose.yml. I think we can make postgres the default for docker compose.

glencbz (changes requested):
Awesome initiative! Can't wait to see what other changes you'll propose :)

Some comments on how we can use docker compose more effectively and a bit of onboarding. We can hold off on further changes until we get on a call and walk through more of the code.

README.md
stonefruit: `ERROR: Cannot find "/app/backend/config/config.json". Have you run "sequelize init"?`

I get this error when running `docker run -it call-home`.

stonefruit: did not see `start:dev`. removed `start-local`.

backend/package.json
stonefruit: did not see `start:dev`. removed `start-local`.

README.md
stonefruit: moved to docker-compose.yml

Dockerfile
stonefruit: `ERROR: Cannot find "/app/backend/config/config.json". Have you run "sequelize init"?`

I get this error when running `docker run -it call-home`.

glencbz: Strange, this should be handled by the `.sequelizerc`. Not sure why this fails locally for me too. Let's remove this line for now, and we'll add a ticket to the backlog to look at this.

glencbz (changes requested):
Looks good, don't know why the `config/config.json` thing is happening, but let's investigate that separately.

Dockerfile
stonefruit: added ticket. we should add a label for non feature related engineering tickets to notion.

stonefruit:
Dont think we need to add anything else to this PR as changes to code are minimal (only the seperate dotenv file).

backend/config/config.json
```diff
@@ -0,0 +1 @@
+{}
```
glencbz, line 1: Could you remove this file too? My bad for not flagging this earlier :(

glencbz (changes requested):
Just need to get rid of the last file

glencbz (approved):

PR #7, opened 2020-10-20
Add instructions for seeding user

stonefruit:
Use `BYPASS_AUTH0_LOGIN=yes` env variable to have instant login.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,39 @@ This is very much in the MVP stage.
 
 Backend runs on port `4000` while Frontend runs on port `3000` and can be run using the instructions below.
 
+### Quick Start
+
+- Get a working copy of `.env`
+- Start Backend:
+
+```bash
+cd backend
+npm i
+npm run start:dev
+```
+
+- Seed database with your phone number (must be real number as SMS is used for logging in)
+
+```
+-- For SQLITE
+INSERT INTO AllowlistEntries("phoneNumber", "role", "destinationCountry","createdAt", "updatedAt")
+VALUES ('+65YOURNUMBER', 'ADMIN', 'SG', date('now'), date('now'));
+
+-- For POSTGRES
+INSERT INTO AllowlistEntries("phoneNumber", "role", "destinationCountry","createdAt", "updatedAt")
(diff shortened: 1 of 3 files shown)


frontend/src/scenes/Login.js
```diff
@@ -98,6 +98,11 @@ export default function Login({ locale }) {
           variant="contained"
           disableElevation
           onClick={() => {
+            // When files are served via Heroku (i.e. staging and production), the base URL and ports are the same
+            // for the backend and frontend. In these cases, the redirect to /auth/login triggers a backend endpoint
+            // and executes the oauth strategy.
+            // In development, frontend and backend use different ports (3000 and 4000), and backend endpoint is not
+            // called.
```
glencbz, line 105: This works as expected if the call is initiated from 4000. Could you amend this to reflect that?

glencbz (changes requested):
I am quite hesitant to use this approach because we can disable all of our authentication with a single misconfiguration.

In my own development, I use my personal facebook and I seed a user into the database. I wonder if there is yet another alternative we can consider.

stonefruit:
Ah ok. Just tried logging in the way you described with Facebook. I think this is the only data that requires seeding right? I will add the sql command in the readme and remove the suggested bypass.

frontend/src/scenes/Login.js
stonefruit: i've put it all in the readme instead

stonefruit:
Reduced PR to just instructions in readme.

README.md
```diff
@@ -70,3 +103,20 @@ Just change the `ports` configuration.
...
+Specifically, the request to GET '/oauth/login' does not work when using React server (:3000)
+
+## Development using Twilio Voice
+
+Twilio requires a public URL to confirm calls via webhook. The webhook URL can be set up with https://ngrok.com/ for now.
+
+In Twilio's console, update the REQUEST URL with your ngrok `https` url.
+`Voice / TwiML / TwiML Apps / Dev backend > REQUEST URL`
+
+Note that this method only works for one developer at a time.
```
glencbz, line 122: We can add a new TwiML app for each dev, then we won't mess with each other's development :)

glencbz (approved):

PR #10, opened 2020-10-21
Send SMS when user added to allowed list

stonefruit:
Dependent on https://github.com/glencbz/call-home/pull/8.

Requires env variable set to 

`SEND_SMS_TO_USER_WHEN_ADDED_TO_ALLOW_LIST=yes`

- [ ] Requires message text to be confirmed


backend/src/services/AllowlistEntry.js
```diff
@@ -15,13 +16,21 @@ function AllowlistEntryService(AllowlistEntryModel) {
...
-    return sanitizeDbErrors(() =>
+    const dbResponse = await sanitizeDbErrors(() =>
       AllowlistEntryModel.create({
         role: UserTypes.USER,
         phoneNumber,
         destinationCountry,
       })
     );
+
+    if (process.env.SEND_SMS_TO_USER_WHEN_ADDED_TO_ALLOW_LIST === 'yes') {
```
glencbz, line 27: The pattern for env vars is that we always destructure them at the top level. This makes it easier to keep track of what vars are needed and what fallbacks we need.

For feature flags, we hide them behind `services/Features` so that we can centralise the logic. This way, we don't have to worry about keeping the flag in sync across the different modules and we can make the logic arbitrarily complex. For example, if we needed to flag more things with this flag, we would need to remember the name of the env var and remember the magic string that we need to set it to.

backend/src/services/AllowlistEntry.js
```diff
@@ -15,13 +16,21 @@ function AllowlistEntryService(AllowlistEntryModel) {
...
+      await sendSms(
+        'Welcome to Call Home. Your account is ready!',
```
glencbz, line 29: This is likely to need configuring, so it should be a static constant at the top.

backend/src/services/AllowlistEntry.js
```diff
@@ -15,13 +16,21 @@ function AllowlistEntryService(AllowlistEntryModel) {
   }
 
   async function createAllowlistEntry({ phoneNumber, destinationCountry }) {
-    return sanitizeDbErrors(() =>
+    const dbResponse = await sanitizeDbErrors(() =>
```
glencbz, line 19: The type of this is allowlistEntry, the name should reflect that.

glencbz (approved):

PR #11, opened 2020-10-28
update tsconfig strictNullChecks:true

stonefruit:
https://github.com/vriad/zod requires strictNullChecks to work. Seems like a good idea to catch such errors too. This PR adds strictNullChecks:true to `tsconfig.json` and adds checks such that the TS compiles.

Needs confirmation if the  logger > rethrow pattern is ok or an alternative should be used.

### DIFF
--- a/backend/src/models/User.ts
+++ b/backend/src/models/User.ts
@@ -51,8 +51,8 @@ class User extends Model<User> {
     },
   })
   @Unique
-  @Column
-  phoneNumber: string;
+  @Column(DataType.TEXT)
+  phoneNumber: string | null;
 
   @Default(false)
   @Column
@@ -61,8 +61,8 @@ class User extends Model<User> {
   @Validate({
     isIn: [['SG', 'BD', '']],
   })
-  @Column
-  destinationCountry: string;
+  @Column(DataType.TEXT)
+  destinationCountry: string | null;
 }
 
 export default User;
--- a/backend/src/routes/Call.ts
+++ b/backend/src/routes/Call.ts
@@ -24,7 +24,9 @@ function CallRoutes(
     );
     const totalDurationSeconds = twilioCalls
       .map((twilioCall) => {
-        if (!twilioCall) return 0;
+        if (!twilioCall) {
+          return 0;
+        }
         return twilioCall.duration;
       })
       .reduce((acc, curr) => acc + curr, 0);
--- a/backend/src/routes/Call.ts
+++ b/backend/src/routes/Call.ts
@@ -23,7 +23,10 @@ function CallRoutes(
       )
     );
     const totalDurationSeconds = twilioCalls
-      .map((twilioCall) => twilioCall.duration)
+      .map((twilioCall) => {
+        if (!twilioCall) return 0;
+        return twilioCall.duration;
+      })
       .reduce((acc, curr) => acc + curr, 0);
     stopwatch.stop();
     req.log.info(
--- a/backend/src/services/AllowlistEntry.ts
+++ b/backend/src/services/AllowlistEntry.ts
@@ -50,8 +50,8 @@ function AllowlistEntryService(
       },
     });
     if (!allowlistEntry) {
-      const msg = 'allowlistEntry not found';
-      logger.error({ id, function: 'deleteAllowlistEntry' }, msg);
+      const msg = `allowlistEntry not found for id ${id}`;
+      logger.error(msg);
       throw new Error(msg);
     }
     await allowlistEntry.destroy();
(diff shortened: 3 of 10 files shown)


backend/src/models/User.ts
```diff
@@ -51,8 +51,8 @@ class User extends Model<User> {
     },
   })
   @Unique
-  @Column
-  phoneNumber: string;
+  @Column(DataType.TEXT)
```
stonefruit, line 54: required for type inference for sequelize to create db as it is ambiguous via typescript

backend/src/routes/Call.ts
```diff
@@ -23,7 +23,10 @@ function CallRoutes(
       )
     );
     const totalDurationSeconds = twilioCalls
-      .map((twilioCall) => twilioCall.duration)
+      .map((twilioCall) => {
+        if (!twilioCall) return 0;
```
stonefruit, line 27: `twilioCall: TwilioCall | null`. In the case of `null`, no need to increment total'

backend/src/services/AllowlistEntry.ts
```diff
@@ -27,7 +28,7 @@ function AllowlistEntryService(
   async function createAllowlistEntry({
     phoneNumber,
     destinationCountry,
-  }: Partial<AllowlistEntryEntity>) {
+  }: Pick<AllowlistEntryEntity, 'phoneNumber' | 'destinationCountry'>) {
```
stonefruit, line 31: Update to `Pick` as `Partial` makes all attributes optional (ie `string | undefined`)

backend/src/models/User.ts
glencbz: Interesting, what do you see? I was certain that enabling `    "emitDecoratorMetadata": true,` in tsconfig was enough for this.

backend/src/routes/Call.ts
glencbz: Is this the style generated by Prettier? I thought prettier + eslint would enforce braces around conditionals

backend/src/services/AllowlistEntry.ts
```diff
@@ -48,9 +49,13 @@ function AllowlistEntryService(
         id,
       },
     });
+    if (!allowlistEntry) {
+      const msg = `allowlistEntry not found for id ${id}`;
```
glencbz, line 53: This wouldn't be 'not found', it would be 'failed to be created'

backend/src/services/TwilioCall.ts
```diff
@@ -114,6 +114,11 @@ class TwilioCallService extends TypedEventEmitter<
     const callEntity = await this.callService.getCallByIncomingSid(
       twilioCall.parentCallSid
     );
+    if (!callEntity) {
+      const msg = `callEntity not found for parentCallSid ${twilioCall.parentCallSid}`;
```
glencbz, line 118: Hm, is this prettier formatting?

glencbz (commented):
Looks good, some comments about one of the log statements and I'm not clear on why the @Column decorator needs to change.

backend/src/models/User.ts
glencbz: Discussed in slack, this is needed because of the new `| null` type, which is needed because we actually assign this property to `null`.

backend/src/routes/Call.ts
stonefruit: duno leh. it didnt format for me. im adding curly manually

backend/src/services/AllowlistEntry.ts
stonefruit: but this is in deleteAllowlistEntry

backend/src/services/TwilioCall.ts
stonefruit: nope. mine.

backend/src/services/AllowlistEntry.ts
glencbz: oh. i'm dumb. resolving this

backend/src/services/TwilioCall.ts
glencbz: Let's put it on one line, don't see a need for this to be on two lines?

stonefruit: what do u mean on one line? it is part of a block of code 
```    if (!callEntity) {
      const msg = `callEntity not found for parentCallSid ${twilioCall.parentCallSid}`;
      logger.error(msg);
      throw new Error(msg);
    }```

stonefruit:
created new squashed PR

backend/src/services/TwilioCall.ts
glencbz: I think I was tired when I was reviewing this, looks like the line break was from github.

PR #13, opened 2020-10-29
Create VerifyFIN.ts

fabianchua6:
added FIN verification

### DIFF
--- a/backend/src/util/VerifyFIN.ts
+++ b/backend/src/util/VerifyFIN.ts
@@ -1,34 +1,32 @@
+/* Sources:
+References supporting the algorithm
+http://www.ngiam.net/NRIC/ppframe.htmSites
+
+To generate and validate FINs
+https://nric.biz
+https://samliew.com/nric-generator
+*/
+
 const checkAlpha = ['X', 'W', 'U', 'T', 'R', 'Q', 'P', 'N', 'M', 'L', 'K'];
 const weightArray = [2, 7, 6, 5, 4, 3, 2];
 
-function validateFIN(fin: string): string {
-  const finUpper = fin.trim.toUpperCase();
-  if (finUpper.length === 9) {
-    const sumArray = fin.slice(1, 8).split('').map(Number); // FIN String into array of numbers to multiply by weight
-
-    let sum = 0;
-    for (let i = 0; i < sumArray.length; i += 1) {
-      sum += sumArray[i] * weightArray[i];
-    } // Cross multiply with weightArray to get sum
-
-    if (finUpper[0] === 'G') {
-      sum += 4;
-    } // FIN that starts with G has their checkAlpha shifted by 4
-
-    if (checkAlpha[sum % 11] === finUpper.slice(-1)) {
-      return true;
-    }
+function validateFIN(fin) {
+  const finUpper = fin.trim().toUpperCase();
+  if (finUpper.length !== 9) {
     return false;
   }
-  return false;
+
+  const sumArray = fin.slice(1, 8).split('').map(Number); // FIN String into array of numbers to multiply by weight
+  let sum = 0;
+  for (let i = 0; i < sumArray.length; i += 1) {
+    sum += sumArray[i] * weightArray[i];
+  } // Cross multiply with weightArray to get sum
+
+  if (finUpper[0] === 'G') {
+    sum += 4;
+  } // FIN that starts with G has their checkAlpha shifted by 4
+
+  return checkAlpha[sum % 11] === finUpper.slice(-1);
 }
 
 export default validateFIN;
-
-// Tests
-// console.log(validateFIN("F1234567N")) // Valid
-// console.log(validateFIN("F1234567A")) // Invalid
-// console.log(validateFIN("G123456T" )) // Not 9 digits.
-// console.log(validateFIN("123456789")) // 9 numbers.
-// console.log(validateFIN("f3725759K")) // Lowercase first alpha
-// console.log(validateFIN("G8977661p")) // Lowercase last alpha
--- a/backend/src/util/VerifyFIN.ts
+++ b/backend/src/util/VerifyFIN.ts
@@ -1,5 +1,5 @@
 /* Sources:
-Public References supporting the algorithm
+References supporting the algorithm
 http://www.ngiam.net/NRIC/ppframe.htmSites
 
 To generate and validate FINs
@@ -10,7 +10,7 @@ https://samliew.com/nric-generator
 const checkAlpha = ['X', 'W', 'U', 'T', 'R', 'Q', 'P', 'N', 'M', 'L', 'K'];
 const weightArray = [2, 7, 6, 5, 4, 3, 2];
 
-function validateFIN(fin: string): boolean {
+function validateFIN(fin) {
   const finUpper = fin.trim().toUpperCase();
   if (finUpper.length !== 9) {
     return false;
--- a/backend/src/util/VerifyFIN.ts
+++ b/backend/src/util/VerifyFIN.ts
@@ -1,6 +1,6 @@
 /* Sources:
-Public References supporting the algorithm
-http://www.ngiam.net/NRIC/ppframe.htmSites
+References supporting the algorithm
+http://www.ngiam.net/NRIC/ppframe.htm
 
 To generate and validate FINs
 https://nric.biz
@@ -16,10 +16,10 @@ function validateFIN(fin: string): boolean {
     return false;
   }
 
-  const sumArray = fin.slice(1, 8).split('').map(Number); // FIN String into array of numbers to multiply by weight
+  const numArray = fin.slice(1, 8).split('').map(Number); // FIN String into array of numbers to multiply by weight
   let sum = 0;
-  for (let i = 0; i < sumArray.length; i += 1) {
-    sum += sumArray[i] * weightArray[i];
+  for (let i = 0; i < numArray.length; i += 1) {
+    sum += numArray[i] * weightArray[i];
   } // Cross multiply with weightArray to get sum
 
   if (finUpper[0] === 'G') {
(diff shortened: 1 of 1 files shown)


backend/src/util/VerifyFIN.ts
```diff
@@ -0,0 +1,34 @@
...
+      return `${fin} is a valid FIN.`;
+    }
+    return `${fin} is an invalid FIN.`;
+  }
+  return 'Wrong Format.';
+}
+
+export default validateFIN;
+
+// Tests
```
stonefruit, line 28: Hey. Are you familiar with jest? would be good to have these guys as unit tests since you already have the scenarios.

backend/src/util/VerifyFIN.ts
```diff
@@ -0,0 +1,34 @@
...
+    for (let i = 0; i < sumArray.length; i += 1) {
+      sum += sumArray[i] * weightArray[i];
+    } // Cross multiply with weightArray to get sum
+
+    if (finUpper[0] === 'G') {
+      sum += 4;
+    } // FIN that starts with G has their checkAlpha shifted by 4
+
+    if (checkAlpha[sum % 11] === finUpper.slice(-1)) {
+      return `${fin} is a valid FIN.`;
```
stonefruit, line 19: Function might be easier to use if it returns a boolean if valid.

backend/src/util/VerifyFIN.ts
```diff
@@ -0,0 +1,34 @@
+const checkAlpha = ['X', 'W', 'U', 'T', 'R', 'Q', 'P', 'N', 'M', 'L', 'K'];
+const weightArray = [2, 7, 6, 5, 4, 3, 2];
+
+function validateFIN(fin: string): string {
+  const finUpper = fin.toUpperCase();
```
stonefruit, line 5: May use `fin.trim().toUpperCase()` in case extra spaces are added.

stonefruit:
Could you leave a link (if there is 1) to the place where you have found the procedure for the FIN verification? thanks.

backend/src/util/VerifyFIN.ts
fabianchua6: Agree!

backend/src/util/VerifyFIN.ts
fabianchua6: Yeap I understand, tbh many of these are still mental placeholders because I'm still not sure how the current stack works. The code is still very raw but the logic is there, feel free to change around with them, appreciate the inputs!

backend/src/util/VerifyFIN.ts
fabianchua6: Not really but I will look it up!

glencbz: Agree. It might be a bit difficult to set up Jest from scratch, but I think it's a good exercise for you @fabianchua6. No pressure, I can come in and help you perfect it. Just try to get the following:

- install jest
- write at least one test case
- run the tests

glencbz:
Agree with @stonefruit. Multiple sources would be preferred.

backend/src/util/VerifyFIN.ts
```diff
@@ -0,0 +1,34 @@
+const checkAlpha = ['X', 'W', 'U', 'T', 'R', 'Q', 'P', 'N', 'M', 'L', 'K'];
+const weightArray = [2, 7, 6, 5, 4, 3, 2];
+
+function validateFIN(fin: string): string {
+  const finUpper = fin.trim.toUpperCase();
+  if (finUpper.length === 9) {
```
glencbz, line 6: Reduce the nesting by returning false if the length is not 9.

backend/src/util/VerifyFIN.ts
```diff
@@ -0,0 +1,34 @@
...
+    let sum = 0;
+    for (let i = 0; i < sumArray.length; i += 1) {
+      sum += sumArray[i] * weightArray[i];
+    } // Cross multiply with weightArray to get sum
+
+    if (finUpper[0] === 'G') {
+      sum += 4;
+    } // FIN that starts with G has their checkAlpha shifted by 4
+
+    if (checkAlpha[sum % 11] === finUpper.slice(-1)) {
```
glencbz, line 18: Collapse this if..return into a single return statement e.g. return actual === expected

glencbz: What is this magic number 11? If it is the length of the checkSum array, use that value instead

backend/src/util/VerifyFIN.ts
```diff
@@ -0,0 +1,34 @@
...
+  const finUpper = fin.trim.toUpperCase();
+  if (finUpper.length === 9) {
+    const sumArray = fin.slice(1, 8).split('').map(Number); // FIN String into array of numbers to multiply by weight
+
+    let sum = 0;
+    for (let i = 0; i < sumArray.length; i += 1) {
+      sum += sumArray[i] * weightArray[i];
+    } // Cross multiply with weightArray to get sum
+
+    if (finUpper[0] === 'G') {
```
glencbz, line 14: I'm wary of this magic constant. What other values can the first letter be? Do they have any special behavior?

fabianchua6:
> Agree with @stonefruit. Multiple sources would be preferred. I'd need to be fairly convinced that this is the right algorithm.
> Could you leave a link (if there is 1) to the place where you have found the procedure for the FIN verification? thanks.

I understand the concerns regarding the validity of the algorithm. Sad to say, my references to the algorithm is only limited to NRIC generators and validators websites like nric.biz and https://samliew.com/nric-generator. These were my main two references in validating this algorithm.

Even though the above sources has managed to successfully reverse engineered the algorithm, I don't know how much I can divulge about it, but I have experiences working with Unique Identification No. validation in the bank where the algorithm is legally obtained from ICA with a license. Therefore, I can confirm that this algorithm is trustable and has been tested multiple times.

backend/src/util/VerifyFIN.ts
fabianchua6: The FIN structure is as such, @XXXXXXX#:
where FIN issued before 1 Jan 2000 are assigned the letter "F" and if it's after 1 Jan 2000, it is are assigned the letter "G". They do have a special behaviour that affects the last character, '#'.

For example, F00000001U and G0000001P will have the same sum when cross multiplying sumArray and weightArray. Therefore, to differentiate between the two, there is a shift of +4 to ensure the last character is different.

backend/src/util/VerifyFIN.ts
fabianchua6: Sorry, could you elaborate? don't quite what get this means :(

backend/src/util/VerifyFIN.ts
fabianchua6: The magic number is part of the algorithm to mod the sum by 11. Not sure why they use 11 too..

backend/src/util/VerifyFIN.ts
glencbz: This is pattern called [early return](https://medium.com/swlh/return-early-pattern-3d18a41bba8). The idea is that code becomes more difficult to understand every time you add another level of `if` or `for` etc (this is called nesting).

We'll look at this function as an example from L19-24, we have these lines:

```js
      return true;
    }
    return false;
  }
  return false;
}
```

At one glance, I can't tell what is what. Does the first return belong to `if (finUpper.length === 9) {`? or is it the second one? And I also have to read past the entire body of the `if` only to realise at the end that I just return false.

To apply this pattern, think of your code as 'exiting' early if the conditions are not met. You want to do all of the heavy lifting at the end of the function instead of making your reviewer read everything up front and then worrying whether or not there is more afterwards.

In this case:

```js
function validateFin(fin) {
  if (fin.length !== 9) {
    return false;
  }
  //..... check the checksum
  return isChecksumValid;
}
```

Notice how we no longer have `if` in `if` and I can immediately find out where the important stuff is.

backend/src/util/VerifyFIN.ts
glencbz: I see. I'll need two things then:

- Cite a source for this
- Capture this information in a code comment so that nobody has to ask this again

glencbz:
That's good enough for me. I just need you to cite these sources in code comments.

backend/src/util/VerifyFIN.ts
```diff
@@ -0,0 +1,34 @@
+const checkAlpha = ['X', 'W', 'U', 'T', 'R', 'Q', 'P', 'N', 'M', 'L', 'K'];
```
glencbz, line 1: At the top of the file, add a big comment block that cites all of your sources

backend/src/util/VerifyFIN.ts
glencbz: Hmm.... Ok I'm comfortable keeping this as just `11`. I'm fairly certain it's the length of the array (because that's a common way of computing checksums), but it's better not to make assumptions.

backend/src/util/VerifyFIN.ts
```diff
@@ -0,0 +1,32 @@
+/* Sources:
+References supporting the algorithm
+http://www.ngiam.net/NRIC/ppframe.htmSites
```
glencbz, line 3: This link is broken

backend/src/util/VerifyFIN.ts
```diff
@@ -0,0 +1,32 @@
...
+const checkAlpha = ['X', 'W', 'U', 'T', 'R', 'Q', 'P', 'N', 'M', 'L', 'K'];
+const weightArray = [2, 7, 6, 5, 4, 3, 2];
+
+function validateFIN(fin) {
+  const finUpper = fin.trim().toUpperCase();
+  if (finUpper.length !== 9) {
+    return false;
+  }
+
+  const sumArray = fin.slice(1, 8).split('').map(Number); // FIN String into array of numbers to multiply by weight
```
glencbz, line 19: This isn't an array of sums, this would be closer to `numArray`.

glencbz (changes requested):
Just fix the broken link

glencbz:
hi slack!

glencbz (approved):

PR #15, opened 2020-10-31
Set up validation. Add AllowlistEntry.ts route input validation.

stonefruit:


backend/package.json
```diff
@@ -47,7 +47,8 @@
     "sequelize": "^5.21.6",
     "sequelize-cli": "^5.5.1",
     "sequelize-typescript": "^1.1.0",
-    "twilio": "^3.42.2"
+    "twilio": "^3.42.2",
+    "zod": "2.0.0-beta.16"
```
stonefruit, line 51: beta recommended by zod. need to update toe v2 later when it is released.

backend/src/custom.d.ts
```diff
@@ -0,0 +1,11 @@
+// Adds additional params to default express.Request type
+declare namespace Express {
+  export interface Request {
+    // In unauthenticated cases, user is not appended to req.
+    // However it is troublesome to check for presence of user object,
+    // hence type is added as a required
+    user: {
```
stonefruit, line 7: adds type for `req.user` globally

backend/src/routes/helpers/validation.ts
```diff
@@ -0,0 +1,9 @@
+import * as z from 'zod';
+
+export const stringToNumberTransformer = z.transformer(
```
stonefruit, line 3: validates input > transforms > validates output of transform.

e.g. will throw error on input 'ABC'

backend/package.json
glencbz: I hate that package.json doesn't allow comments

backend/src/custom.d.ts
glencbz: There's a type in middlewares called `CallHomeRequest` for this reason. Would that be good enough?

backend/src/routes/helpers/validation.ts
```diff
@@ -0,0 +1,9 @@
...
+  z.string().nonempty(),
+  z.number(),
+  (val) => {
+    return +val;
```
glencbz, line 7: I prefer using explicit `Number` conversion when converting between types.

backend/src/routes/helpers/validation.ts
glencbz: Nit: I generally prefer to export at the end of the module, I think it makes exports easier to track. If you have a good reason for exporting a const, then feel free to ignore.

backend/src/routes/AllowlistEntry.ts
```diff
@@ -17,21 +20,46 @@ function AllowlistEntryRoutes(allowlistService: typeof AllowlistEntry) {
...
+      try {
+        const bodySchema = z.object({
+          phoneNumber: z.string(),
+          destinationCountry: z.string(),
+        });
+        const body = bodySchema.parse(req.body);
+        validatedReq = { body };
+      } catch (error) {
+        logger.error(error);
+        return res.status(400).send(error);
```
glencbz, line 33: This validation looks solid, but there's a lot of duplication between defining the schema and returning the 400 + logging the error. Do you think we can refactor this to reduce the duplication and make it cleaner?

glencbz: Here's an example i was able to come up with:

```js
// helper function file
function validateRequest<T>(schema: z.Schema<T>, handler: (parsedReq: T, res: Response) => unknown) {
  return (req: Request, res: Response) => {
    try {
      const parsedReq = schema.parse(req);
      return handler(parsedReq, res);
    } catch (error) {
      res.status(400).send(error);
    }
  }
}
```

```js
// In the route file
const POST_SCHEMA = z.object({
  body: z.object({
    phoneNumber: z.string(),
    destinationCountry: z.string(),
  })
});

  router.post('/', requireAdmin, validateRequest(POST_SCHEMA, async (req, res) => {
    try {
      const { phoneNumber, destinationCountry } = req.body;
      const allowlistEntry = await allowlistService.createAllowlistEntry({
        phoneNumber,
        destinationCountry,
      });
      return res.status(200).json(allowlistEntry);
    } catch (e) {
      return handleServiceError(e, res);
    }
  }));


```

stonefruit: This would mean that `phoneNumber` would be type `any` is it? I wanted to try using a helper function too, but could not figure out how to preserve the types, which is pretty nice to have directly linked to the validator which is what zod provides. Not sure if the reduction in repetition of code is required compared to keeping it straightforward.

glencbz: My concerns with not helping a helper are that duplication is error prone, and we have to introduce quite a few nested blocks + return statements to preserve the correct order of operations. It feels a little too error-prone at the moment. Let me push a PR with a better implementation.

backend/src/custom.d.ts
stonefruit: yea i guess so, just that we will have to explicitly give the type each time req appears. allows more granular control of the type though. lets continue with using just `CallHomeRequest`

backend/src/routes/helpers/validation.ts
stonefruit: Will use Number in subsequent PRs

backend/src/routes/helpers/validation.ts
stonefruit: ok

backend/src/custom.d.ts
glencbz: Agreed, perhaps we should just use `CallHomeRequest` for these route handlers.

The reason why I made this two different types is because `req.user` is only included under certain conditions, so making them distinct in the code might pay dividends. Unfortunately, I can't figure out a way to make express figure out the types via the middlewares.

glencbz (approved):

PR #17, opened 2020-11-01
Add jest

fabianchua6:
Installing jest and also configuring it to work with typescript


glencbz (changes requested):
There doesn't seem to be any jest here

stonefruit:
hmm. is this just adding jest? or also the test for the FIN ?

fabianchua6:
Yes it's only for adding jest, without the test for the fin. Should I include it in too?

backend/src/util/VerifyFIN.test.ts
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable */
```
glencbz, line 1: Why `eslint-disable`?

backend/src/util/VerifyFIN.test.ts
```diff
@@ -0,0 +1,30 @@
...
+import validateFIN from "./VerifyFIN"
+
+test('Valid FIN', () => {
+    expect(validateFIN("F1234567N")).toBeTruthy
+})
+
+test('Lowercase first alphabet', () => {
+    expect(validateFIN("f3725759K")).toBeTruthy
```
glencbz, line 9: The assertions need to be called, e.g. `toBeTruthy()`.

backend/src/util/VerifyFIN.test.ts
fabianchua6: Thought it was a workaround for using tests files but I managed to find another method which adds jest env to eslint, makes much more sense I believe. Thanks for pointing this out!

backend/src/util/VerifyFIN.test.ts
fabianchua6: Got it!

glencbz (approved):
Awesome stuff. Remember to resolve the merge conflicts.

fabianchua6:
Finally... that felt good

glencbz (changes requested):
Did you delete frontend/package-lock.json?

glencbz:
There were merge conflicts, so I fixed this out-of-band. Good work @fabianchua6!

PR #18, opened 2020-11-04
Add input validation to passwordless

stonefruit:


backend/src/routes/AllowlistEntry.ts
```diff
@@ -36,9 +36,9 @@ function AllowlistEntryRoutes(allowlistService: typeof AllowlistEntry): Router {
   router.post(
     '/',
     requireAdmin,
-    validateRequest(POST_SCHEMA, async (req, res) => {
+    validateRequest(POST_SCHEMA, async (parsedReq, res) => {
```
stonefruit, line 39: updated for consistency

backend/src/routes/helpers/validation.ts
```diff
@@ -10,11 +11,15 @@ type RequestSchema<Output, TD extends z.ZodTypeDef> = z.Schema<
...
+    parsedReq: Output,
+    res: Response,
+    rawReq: CallHomeRequest
+  ) => unknown
 ) {
-  return async (req: Request, res: Response) => {
+  return async (req: CallHomeRequest, res: Response) => {
     try {
-      const parsedReq = schema.parse(req as Partial<Request>);
+      const parsedReq = schema.parse(req as Partial<CallHomeRequest>);
```
stonefruit, line 22: `schema.parse` will remove types not in the schema. Hence, `parsedReq` will only contain the stuff defined in the schema.

The `req` in the 3rd return value will need to be used if stuff like `req.log.error` is to be used in the router code.

stonefruit:
View first commit to verify js > ts conversion.

backend/src/routes/Passwordless.ts
```diff
@@ -0,0 +1,109 @@
...
+  passwordlessRequestService: typeof PasswordlessRequest
+): Router {
+  const router = express.Router();
+
+  router.post(
+    '/begin',
+    validateRequest(BEGIN_SCHEMA, async (parsedReq, res, req) => {
+      const { id } = req.user;
+      if (!id) {
+        req.log.error('wtf something is going wrong');
```
glencbz, line 34: lol

backend/src/routes/Passwordless.ts
```diff
@@ -0,0 +1,109 @@
...
+      }
+      const passwordlessRequests = await passwordlessRequestService.getPasswordlessRequestsByUserId(
+        id
+      );
+      const lastRequest = passwordlessRequests[passwordlessRequests.length - 1];
+      if (
+        lastRequest &&
+        Number(new Date()) - lastRequest.requestTime <=
```
glencbz, line 42: Nit: Hm, it is a little surprising that this works, but I can verify that it does indeed work. Maybe we should be consistent and cast both to `Number` just for consistency?

backend/src/routes/helpers/validation.ts
glencbz: Yup. That was the intent.

backend/src/routes/Passwordless.ts
```diff
@@ -0,0 +1,109 @@
...
+  userService: typeof User,
+  auth0Service: typeof Auth0,
+  passwordlessRequestService: typeof PasswordlessRequest
+): Router {
+  const router = express.Router();
+
+  router.post(
+    '/begin',
+    validateRequest(BEGIN_SCHEMA, async (parsedReq, res, req) => {
+      const { id } = req.user;
```
glencbz, line 32: I think we should add `req.user` this to the schema, what do you think?

glencbz (commented):
Nice! This is good stuff. I'm thinking of including `req.user` in the schema, since it is technically part of the input. What do you think?

backend/src/routes/Passwordless.ts
stonefruit: Seems like different things though. the schema is meant for stuff coming from the client, whereas user is coming from the middleware. like someone might expect user.id to be somewhere in the request sent from the browser if they didnt know.

backend/src/routes/Passwordless.ts
stonefruit: Sure. requestTime is a Date type in the model so it also returns a javascript date.

backend/src/routes/Passwordless.ts
glencbz: I see, I think we're looking at this two different ways. You're thinking of using the schema to guard against client behavior and I'm thinking of using the schema to give type safety to route handlers. I'm ok with sticking with your approach, it sounds reasonable. I will add zod validation to the middleware too so that we can have safety for `req.user`.

glencbz (approved):

PR #19, opened 2020-11-10
Add input validation for transaction

stonefruit:


backend/src/routes/Transaction.ts
```diff
@@ -1,43 +1,49 @@
...
-        return res
-          .status(400)
-          .json('Invalid transaction, amount must be specified');
-      }
-      if (!Number(userId) || !Number(amount)) {
-        return res.status(400).json('Invalid transaction.');
-      }
+    validateRequest(POST_SCHEMA, async (parsedReq, res) => {
+      const { userId } = parsedReq.params;
+      const { amount } = parsedReq.body;
```
glencbz, line 39: Nit: since we have high confidence that this is parsed correctly, we can do a single nested destructuring instead of putting this on two lines. I don't really have a preference, just pointing it out :)

backend/src/routes/helpers/validation.ts
```diff
@@ -29,4 +29,9 @@ function validateRequest<Output, TD extends z.ZodTypeDef, Input>(
   };
 }
 
-export { validateRequest };
+const stringToNumberTransformer = z
```
glencbz, line 32: Cleanup task after this is merged is to reuse this helper in the AllowlistEntry routes.

glencbz (approved):

PR #26, opened 2020-11-29
Add call minutes progress bar

xiendong:
Here's a PR that updates the call limits with a progress bar

image

image
Full bar

image
Empty bar

Thanks!

### DIFF
--- a/frontend/src/scenes/ContactsPage.tsx
+++ b/frontend/src/scenes/ContactsPage.tsx
@@ -13,8 +13,7 @@ import CallIcon from '@material-ui/icons/Call';
 import PhoneDisabledIcon from '@material-ui/icons/PhoneDisabled';
 import ExitToAppIcon from '@material-ui/icons/ExitToApp';
 import FeedbackIcon from '@material-ui/icons/Feedback';
-import PhoneInTalkIcon from '@material-ui/icons/PhoneInTalk';
-import AutorenewIcon from '@material-ui/icons/Autorenew';
+import RefreshIcon from '@material-ui/icons/Refresh';
 import FavoriteIcon from '@material-ui/icons/Favorite';
 import HistoryIcon from '@material-ui/icons/History';
 import ArrowForwardIcon from '@material-ui/icons/ArrowForward';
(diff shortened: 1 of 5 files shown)


frontend/src/scenes/ContactsPage.tsx
```diff
@@ -438,46 +444,71 @@ function EditContactDialog({
...
-      <InfoItem variant="subtitle2" className="info-item">
-        <FavoriteIcon className="info-icon" />
-        {contacts.length} {STRINGS[locale].CONTACTS_LOVED_ONES_LABEL}
-      </InfoItem>
+      <div
+        className="info-container"
+        style={{ justifyContent: 'space-between', marginTop: 4 }}
+      >
+        <CallTimeInfoItem variant="body2">
+          {formatDurationInMinutes(userCallTimeDuration)} left
```
glencbz, line 499: Try to avoid putting literal strings in the template directly. It will affect localisation.

frontend/src/scenes/ContactsPage.tsx
```diff
@@ -438,46 +444,71 @@ function EditContactDialog({
...
+        </CallTimeInfoItem>
+
+        {nextRefreshTimeString && nextRefreshAmount && (
+          <LightInfoItem
+            variant="body2"
+            style={{ display: 'flex', alignItems: 'center' }}
+          >
+            <RefreshIcon style={{ marginRight: 4 }} />+{nextRefreshAmount} in{' '}
```
glencbz, line 507: Try to avoid putting literal strings in the template directly. It will affect localisation.

glencbz (changes requested):
Awesome stuff! Would prefer to get the time amounts and units from the backend, but I understand that it would be add a fair amount of work to your plate. I think it'll be good learning though, let me know if you're comfortable with that.

frontend/src/scenes/ContactsPage.tsx
xiendong: @glencbz hmm probably can put the appropriate strings to the localisation dictionary object, but how to obtain the translated string?

xiendong:
Addressed the comments

frontend/src/scenes/ContactsPage.tsx
```diff
@@ -83,18 +83,33 @@ const EN_STRINGS = {
   CONTACTS_REACHED_CALL_LIMIT_MESSAGE:
     'You have reached your call limit for the month',
   CONTACTS_TIME_TO_CREDIT_MESSAGE: function CreditMessage(
-    timeAmount: string | null,
-    creditAmount: string | null
+    timeAmount: Duration | null,
+    creditAmount: Duration | null
   ) {
-    if (!timeAmount) {
+    if (timeAmount == null || creditAmount == null) {
```
glencbz, line 89: Use `===`

glencbz (approved):

PR #29, opened 2020-12-08
update recent calls to use Twilio parent call Sids

stonefruit:


frontend/src/scenes/ContactsPage.tsx
```diff
@@ -721,14 +720,12 @@ export default function ContactsPage({ locale, routePath }: SceneProps) {
...
         </Typography>
-        {false && (
-          <Link to={PATHS.RECENT_CALLS} style={{ textDecoration: 'none' }}>
-            <ActionLink variant="body1" role="button">
-              <HistoryIcon />
-              {STRINGS[locale].CONTACTS_RECENT_CALLS}
-            </ActionLink>
-          </Link>
-        )}
+        <Link to={PATHS.RECENT_CALLS} style={{ textDecoration: 'none' }}>
```
stonefruit, line 723: only this `false` was removed. other changes in file seem to be from previously unlinted file.

glencbz (approved):

PR #34, opened 2020-12-15
update/login page updates

fabianchua6:
Updating the login page cover image and added facebook messenger support.


frontend/src/scenes/Login.js
```diff
@@ -90,6 +102,10 @@ export default function Login({ locale, routePath }) {
         >
           {STRINGS[locale].FACEBOOK_SIGN_UP}
         </FacebookButton>
+        <ActionLink variant="body1" role="button" onClick={()=>{window.location.href='https://m.me/callhomesg'}}>
```
glencbz, line 105: Use `<a>` and `href` for simple navigation. You can remove the `ActionLink` component if you do.

glencbz (changes requested):
Looks good, just some minor tweaks

frontend/src/scenes/Login.js
```diff
@@ -90,6 +94,15 @@ export default function Login({ locale, routePath }) {
         >
           {STRINGS[locale].FACEBOOK_SIGN_UP}
         </FacebookButton>
+        <Link
```
stonefruit, line 97: I think glen meant to use
```
<a href="https://m.me/callhomesg" style="...">the icon and text</a>
```
'<Link>' is used for linking to react router routes within the app. `<a>` uses normal html sementics

glencbz: I'm somewhat confident that <Link> behaves like a <a> for routes that live outside of the application. I can test the changes to confirm this.

glencbz: Link works just fine. Let's standardise on using <Link> so that we don't have to worry about choosing <a> or <Link>

PR #38, opened 2021-02-22
Adding redemption status and user_createdAt timestamp

lyejunting:
- Added timestamp for when a user is signed up. This is for us to sort by the date they signed up with us for a quick count on how many we managed to onboard (without accessing the DB)

- Added a status for the redemption promo codes. Technically promo codes should only show if they are "redeemable" but this can be easily confused. Added for clarity that the redemption count is at 1.

### DIFF
--- a/backend/src/models/User.ts
+++ b/backend/src/models/User.ts
@@ -49,6 +49,9 @@ class User extends Model<User> {
   })
   @Column
   destinationCountry: string;
+
+  @Column
+  createdAt: string;
 }
 
 export default User;
--- a/frontend/src/scenes/AdminPage/UserTabContent.tsx
+++ b/frontend/src/scenes/AdminPage/UserTabContent.tsx
@@ -14,6 +14,7 @@ interface UserGridRow {
   phoneNumber: string | null;
   destinationCountry: string;
   callTime: string | null;
+  createdAt: string | null;
   hasAdminGrantedValidation: boolean;
 }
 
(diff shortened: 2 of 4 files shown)


frontend/src/scenes/AdminPage/UserTabContent.tsx
```diff
@@ -72,9 +73,12 @@ export default function UserTabContent() {
     phoneNumber: user.phoneNumber,
     destinationCountry: user.destinationCountry,
     callTime:
-      user.callTime === undefined || user.callTime === null
+      user.callTime == null
```
lyejunting, line 76: changed this to == to use the abstract equality operator which should cover both (without strict type).

Let me know if you're uncomfortable with this :)

backend/src/models/User.ts
```diff
@@ -49,6 +49,9 @@ class User extends Model<User> {
   })
   @Column
   destinationCountry: string;
+
+  @Column
+  createdAt: string;
```
glencbz, line 54: This column is provided actually by sequelize by default. I believe you can omit this.

frontend/src/scenes/AdminPage/UserTabContent.tsx
glencbz: Yes, I would prefer to use `===` in all cases, even if `==` is appropriate here. Even if all devs understood the behaviour of `==` in every single case, it is too visually similar to `===`. I would prefer to use only `===` so that readers of the code can be confident that it is always `===`.

PR #44, opened 2021-07-08
Regenerate lockfiles and adding Dockerfiles

chaychoong:
### Fixes
- `docker-compose up` should now run the application stack smoothly
- proxying from `localhost:3000/oauth` should now work properly. should not need to run on `localhost:4000` any longer but the option still exists

### Known issues / won't fix
- Seeding and migrating runs every time we `docker-compose up`


backend/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN apt update && apt install -y sqlite3
+
+WORKDIR /app/backend
+
+RUN npm install -g node-gyp node-pre-gyp
+RUN npm install --build-from-source --sqlite=/usr/local sqlite3
+
+COPY package.json .
+COPY package-lock.json .
+RUN npm run init
```
glencbz, line 12: I'm surprised that this works

docker-compose.yml
```diff
@@ -23,5 +12,20 @@ services:
     ports:
       - 5432:5432
 
+  shared:
+    image: call-home-shared
```
glencbz, line 16: What does this image + container do?

backend/Dockerfile
glencbz: Nvm, just saw the base image.

backend/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM call-home-shared
```
glencbz, line 1: Is this name documented anywhere other than docker compose? If not, then this dockerfile isn't usable outside of docker compose. Might be worth creating an npm script in shared/ called `dockerize` that builds and tags the image accordingly.

frontend/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+COPY package-lock.json .
+
+RUN npm run init
+RUN npm install
+
+COPY tsconfig.json .
+COPY src/ src/
+COPY public/ public/
+
+RUN npm start
```
glencbz, line 15: What does this dockerfile do exactly? It looks like this just for dev because this uses the react-scripts build. If it is just for dev, perhaps this should be Dockerfile.dev and the prod one would be Dockerfile.

I think the typical approach for most SPAs in docker is to build the bundle and then serve it with nginx or something similar. I don't think we care enough to serve with nginx, but we can copy the bundle out of the prod image.

glencbz (commented):
Good changes, but I have some open questions on the intent of the dockerfiles

docker-compose.yml
glencbz: It looks like this is build-only?

glencbz:
I was not able to reproduce the issue with the frontend package lock, but I was able to address the sequelize issues and the new typechecking issue.

glencbz:
I've pushed this branch to glencbz:regenerate-lockfiles. I've added you as a collaborator, so you can update that branch too. Could you change the PR to use that branch when you can?

backend/Dockerfile
chaychoong: good idea, will do that

frontend/Dockerfile
chaychoong: just for dev, will rename

docker-compose.yml
chaychoong: yes the intent is to build using this step. will change to building it with an npm target

glencbz: Actually I think build only is fine. We can have both the script and this build only service

frontend/src/setupProxy.js
```diff
@@ -0,0 +1,13 @@
+const createProxyMiddleware = require('http-proxy-middleware');
+
+const { BACKEND_HOST = 'localhost' } = process.env;
+
+module.exports = function (app) {
+  app.use(
+    '/oauth',
```
glencbz, line 7: The redirection after the OAuth doesn't seem to work for me. I get redirected to localhost:4000. We can clean this up later, I don't think this is a blocker.

glencbz (approved):
I've pushed changes to my branch. Once you merge them in, feel free to squash and merge.

There's an outstanding issue with the OAuth redirection, but I think this is a step in the right direction. Let's merge this and fix the issues in a subsequent PR.

PR #45, opened 2021-07-28
Periodically invalidate PhoneNumberValidations

glencbz:
This adds an expiry to PhoneNumberValidations, which means that users will have
to periodically revalidate their phone numbers. This helps ensure that our users
are indeed based in Singapore.

### DIFF
--- a/backend/src/jobs/refreshPhoneNumberStatus.ts
+++ b/backend/src/jobs/refreshPhoneNumberStatus.ts
@@ -1,0 +1,23 @@
+import { logger } from '../config';
+import type { PhoneNumberValidation } from '../services';
+
+// Hourly
+const JOB_INTERVAL_MILLIS = 60 * 60 * 1000;
+
+function refreshPhoneNumberStatus(
+  phoneNumberValidationService: typeof PhoneNumberValidation
+): void {
+  async function job(): Promise<void> {
+    try {
+      logger.info('refreshPhoneNumberStatus==========');
+      await phoneNumberValidationService.invalidateExpiredEntries();
+    } catch (error) {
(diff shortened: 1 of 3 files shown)


backend/src/jobs/refreshPhoneNumberStatus.ts
```diff
@@ -0,0 +1,23 @@
+import { logger } from '../config';
+import type { PhoneNumberValidation } from '../services';
+
+// Hourly
+const jobIntervalMillis = 60 * 60 * 1000;
```
chaychoong, line 5: nit: caps for const

backend/src/services/PhoneNumberValidation.ts
```diff
@@ -65,10 +72,32 @@ function PhoneNumberValidationService(
...
+          // Less than because the expired entries are earlier than our cutoff.
+          [Op.lt]: expiryDate,
+        },
+      },
+    });
+
+    // Expire these in batches
+    await Promise.all(
+      expiredEntries
+        .slice(0, EXPIRY_BATCH_SIZE)
```
chaychoong, line 89: would this just operate on the first 100?

glencbz: Yes, the intent is to only operate on the first 100. Considering that there is a backlog of expired entries, I think that if we try to do everything at once, we'll flood the app with too much work.

backend/src/jobs/refreshPhoneNumberStatus.ts
glencbz: Done

chaychoong (approved):

PR #51, opened 2021-08-31
Call Feedback Dialog Implementation

stevahnes:


glencbz:
Quick comment on the commit message: it is recommended to use the 'imperative mood' in commit messages (see https://chris.beams.io/posts/git-commit/). This is recommended even in the git man page.

frontend/src/scenes/CallingPage.tsx
```diff
@@ -241,10 +256,15 @@ export default function CallingPage({ locale, routePath }: SceneProps) {
...
-    contactService?.setActiveContact(null);
+  const handleExitCallButton = useCallback(() => {
+    if (!hasUserDisconnected) {
+      disconnectCall();
+    }
+    if (activeConnection) {
+      openFeedbackDialog();
+    } else {
+      exitCallingPage();
+    }
```
glencbz, line 267: Comment on the commit message: Even if the behaviour can be inferred from the code, it is usually helpful to describe:

- the problem you are solving
- the behaviour of your code that solves the problem
- breaking changes
- possible things to watch out for
- possible contentious things in the commit

This makes it a lot easier for the reviewer because the reviewer needs to know the answers to a lot of questions, including:

- what does your code do?
- do we want the code to do this?
- is the problem solved?

frontend/src/scenes/CallingPage.tsx
```diff
@@ -241,10 +256,15 @@ export default function CallingPage({ locale, routePath }: SceneProps) {
     setHasUserDisconnected(true);
   }, [activeConnection]);
 
-  const exitCallingPage = useCallback(() => {
-    disconnectCall();
-    userService?.setShouldSleep(true);
-    contactService?.setActiveContact(null);
+  const handleExitCallButton = useCallback(() => {
+    if (!hasUserDisconnected) {
+      disconnectCall();
```
glencbz, line 261: The original callback calls `disconnectCall()` unconditionally. Is there a particular reason why you added this `if`?

frontend/src/scenes/CallingPage.tsx
```diff
@@ -241,10 +256,15 @@ export default function CallingPage({ locale, routePath }: SceneProps) {
...
+    }
+    if (activeConnection) {
+      openFeedbackDialog();
+    } else {
```
glencbz, line 265: Nit: Preferred style for this codebase is to use early return instead of if-else.

frontend/src/scenes/CallingPage.tsx
```diff
@@ -102,6 +103,21 @@ export default function CallingPage({ locale, routePath }: SceneProps) {
   const { me: user } = userState || {};
   const [contactState, contactService] = useContactService();
   const { activeContact } = contactState || {};
+  const [isFeedbackDialogOpen, setIsFeedbackDialogOpen] = useState(false);
+
+  const exitCallingPage = () => {
+    userService?.setShouldSleep(true);
```
glencbz, line 109: This should probably be moved into `handleExitCallButton()`. We want to prevent sleeping only when a call is active.

frontend/src/components/shared/FeedbackDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { Locale } from 'scenes/types';
+import { Connection } from 'twilio-client';
+import { PrimaryButton } from './RoundedButton';
+
+interface FeedbackDialogProps extends DialogProps {
+  connection: Connection;
+  locale: Locale;
+}
+
+// Reference for Problem Keys: https://www.twilio.com/docs/voice/sdks/javascript/overview-1x-deprecated/connection
```
glencbz, line 18: Great job linking to the docs!

frontend/src/components/shared/FeedbackDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+    GOOD: 'GOOD',
+    BAD: 'BAD',
+  },
+  PROBLEM_TITLES: {
+    'Call got disconnected': Connection.FeedbackIssue.DroppedCall,
+    'Lag in audio': Connection.FeedbackIssue.AudioLatency,
+    "Can't hear one another": Connection.FeedbackIssue.OneWayAudio,
+    'Sound breaking up': Connection.FeedbackIssue.ChoppyAudio,
+    'Poor audio quality': Connection.FeedbackIssue.NoisyCall,
+    'Audible echos during call': Connection.FeedbackIssue.Echo,
```
glencbz, line 32: I wonder if this is understandable to users. I'm not saying we should change it, but we can keep an eye out.

frontend/src/components/shared/FeedbackDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+export default function FeedbackDialog({
+  connection,
+  locale,
+  open,
+  onClose,
+}: FeedbackDialogProps) {
+  const [feedback, setFeedback] = useState(-1);
+  const [hover, setHover] = useState(-1);
+  const [problem, setProblem] = useState<Connection.FeedbackIssue | undefined>(
+    Connection.FeedbackIssue.DroppedCall
```
glencbz, line 63: Why is the default DroppedCall instead of undefined?

frontend/src/components/shared/FeedbackDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+const DialogPrimaryButton = withDialogButtonStyles(PrimaryButton);
+
+export default function FeedbackDialog({
+  connection,
+  locale,
+  open,
+  onClose,
+}: FeedbackDialogProps) {
+  const [feedback, setFeedback] = useState(-1);
+  const [hover, setHover] = useState(-1);
```
glencbz, line 61: We set `hover`, but where is `hover` read?

frontend/src/components/shared/FeedbackDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+            precision={1}
+            onChange={(_event, newValue) =>
+              newValue ? setFeedback(newValue) : setFeedback(-1)
+            }
+            onChangeActive={(_event, newHover) => {
+              setHover(newHover);
+            }}
+          />
+        </div>
+        {feedback < 5 && feedback > 0 ? (
```
glencbz, line 108: My preference is not to use magic values like `-1` because it leads to unusual-looking conditions like this. A reader seeing this might wonder 'why this range?' or 'what other values can this value be?'. In fact, I also think that it has also obscured an off-by-one error (did you mean `>= 0` instead of `> 0`?).

This might be made clearer by making feedback a `number | null` and representing a missing value by `null`. This check would then be `feedback !== null & feedback < 5`, which reads in a natural way - if feedback has a specified value and it is less than perfect, we ask for a reason.

glencbz (changes requested):
Looks good on the whole, only some very very minor tweaks requested.

frontend/src/scenes/CallingPage.tsx
stevahnes: Oh! Okay :)

frontend/src/scenes/CallingPage.tsx
stevahnes: hmmmm, I can't rmb but relooking at the code yea I think better be safe and call `disconnectCall()` unconditionally.

frontend/src/scenes/CallingPage.tsx
stevahnes: Gotcha!

frontend/src/components/shared/FeedbackDialog.tsx
stevahnes: React complained that we are making an uncontrolled variable controlled when setting it to undefined at the start. Unless you are okay with it?

frontend/src/components/shared/FeedbackDialog.tsx
stevahnes: Yeah the copywriting on the left side can be improved haha. Do you have better ideas?

frontend/src/components/shared/FeedbackDialog.tsx
stevahnes: Actually, it is not read because we didn't implement the corresponding labelling to the score (e.g. 1 is poor, 2 is bad, etc). But the sample doc has it. Should I keep it or remove it?

frontend/src/components/shared/FeedbackDialog.tsx
stevahnes: Ahhh I see! Again I took reference from the material UI documentation haha. Think if we set it to `null` we'll also get the warning of making an uncontrolled variable controlled. Wdyt?

frontend/src/scenes/CallingPage.tsx
stevahnes: Tested and it's working fine :)

frontend/src/scenes/CallingPage.tsx
glencbz: Sorry, could you draw the link a little more explicitly for me? I don't see how the code you attached ties in with moving around the `setShouldSleep()` call.

That said, upon reading this again, I realise that your implementation preserves the existing behaviour (we only reenable sleeping when the user exits the page, not when the call is disconnected). I don't believe in asking you to fix current behaviour while you're adding a new feature, so feel free to ignore the code comments :)

frontend/src/components/shared/FeedbackDialog.tsx
glencbz: No, I don't really have better ideas. I think your suggestions are pretty reasonable. Let's just keep an eye on this.

frontend/src/components/shared/FeedbackDialog.tsx
glencbz: Hm, I don't see why this would be a React issue. In either case, we are controlling the form. Could you try setting it to `undefined` and show me the warning/error output that you're seeing?

frontend/src/components/shared/FeedbackDialog.tsx
glencbz: If it is not read, let's remove it.

frontend/src/components/shared/FeedbackDialog.tsx
glencbz: If there are no warnings, I think `null` is an easy decision. If there are warnings, the magic value `-1` is ok, but we should document it. To make sure we're on the same page, could you show me the warning?

frontend/src/scenes/CallingPage.tsx
stevahnes: Okies! Haha... Oh I think I was under the impression that if you set user to sleep, user might become `null` but then realized it's a misunderstanding on my end. Think can try moving it to `handleExitCall()` to make it cleaner.

frontend/src/components/shared/FeedbackDialog.tsx
stevahnes: image

frontend/src/components/shared/FeedbackDialog.tsx
stevahnes: Hmmm, looks like null is fine for the Ratings value 😄

frontend/src/components/shared/FeedbackDialog.tsx
stevahnes: It still works fine tho, despite the warning

glencbz: It looks like you found a good solution here, which is to use `null` instead of `undefined`. Then there isn't even a need to cast, which is great :)

glencbz (approved):
Great stuff, man! Let me get this into prod ASAP

stevahnes:
Ready for more HAHA

PR #67, opened 2022-10-11
chore: tooling improvements

chaychoong:
## Changes
- refactor npm run commands and `./buildAndDeploy.sh` into a Makefile
- add .npmrc files to specify local .npm folder for links and caches
- add flake.nix file so nix users have a standard build env
- add .talismanrc in case anyone uses [talisman](https://github.com/thoughtworks/talisman)


.talismanrc
```diff
@@ -0,0 +1,4 @@
+fileignoreconfig:
+- filename: flake.lock
+  checksum: 9f9722d0a14d372769015f1c57abd64f4bc050ccba75913bc299bab14f6f58fb
+version: ""
```
glencbz, line 4: Unless you are also proposing that talisman should be part of the stack (which would require a lot more legwork) can we remove this?

package.json
```diff
@@ -2,20 +2,6 @@
...
-    "init": "npm run init:shared && npm run init:backend && npm run init:frontend",
-    "init:shared": "cd shared && npm run init",
-    "init:frontend": "cd frontend && npm run init",
-    "init:backend": "cd backend && npm run init",
-    "precommit:shared": "cd shared && npm run precommit",
-    "precommit:frontend": "cd frontend && npm run precommit",
-    "precommit:backend": "cd backend && npm run precommit",
-    "precommit": "npm run precommit:frontend && npm run precommit:backend && npm run precommit:shared"
-  },
   "repository": {
```
glencbz, line 5: All of the scripts have been removed and the new ones have no documentation. Could you add documentation so that devs whose workflows have to change (e.g. me 😛 ) know what they should be using?

glencbz (changes requested):
Thanks for the changes! I really appreciate the effort.

I want to like this change, since I've thought of doing some of these things, but this PR is quite lacking in documentation, and as such it seems under-motivated to me and it's not obvious to me why the changes are good and safe.

- What does the .npmrc do? How does this make our lives better?
- Why the .talismanrc? Do we expect to use talisman?
- What is .direnv doing in .gitignore? If that's a tool you use, shouldn't that be in your global gitignore?
- What value does the makefile give us? How do I use it?

The no. 1 thing that will help this get merged is to document the rationale behind each of these changes in the commit messages, only then will I understand why this is a good change to make.

.talismanrc
chaychoong: removed!

package.json
chaychoong: Modified the README.md and added some documentation to the Makefile,

chaychoong:
> The no. 1 thing that will help this get merged is to document the rationale behind each of these changes in the commit messages, only then will I understand why this is a good change to make.

Added much better commit messages

.gitignore
```diff
@@ -1 +1,2 @@
-node_modules/
\ No newline at end of file
+node_modules/
+.npm/
```
glencbz, line 3: Optional: you could re-add the trailing newline to make this diff less odd-looking. Trailing newlines are expected by Unix anyway https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

glencbz (changes requested):
Nice, thanks for the commit messages, this looks great now.

Do remove `.direnv/`. unless there's a good reason for it to be there, in which case, please add the reason to the commit message.

.gitignore
glencbz: Oh, you _added_ the newline. My b

chaychoong: Of course, auto newline in emacs man

.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules/
+.direnv/
```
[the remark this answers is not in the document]
chaychoong: hmm yes this should be in the user's global gitignore I guess

glencbz (approved):
LGTM. This is some top tier stuff, thanks!

.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules/
 .npm/
```
glencbz, line 2: Optional, but preferred: could you drop this merge commit and rebase the branch instead?

Makefile
```diff
@@ -0,0 +1,90 @@
...
+APPS:=shared frontend backend
+
+.DEFAULT_GOAL:=help
+
+##@ Dev
+
+init: $(APPS:%=%.init) ## Setup repo, and npm ci
+
+$(APPS:%=%.init): %.init:
+	cd $* && npm run init && npm ci
```
glencbz, line 27: Since `shared/` is a shared library, there's a dependency on `shared.init` in both `backend.init` and `frontend.init`, but these are peers in this rule. Technically not a problem if not running with concurrency, though not ideal. Maybe leave a TODO comment and we can come back to this later.

.gitignore
chaychoong: done

Makefile
chaychoong: done

glencbz:
brrrrilliant

PR #68, opened 2022-10-19
Add untracked file, fix .gitignore

glencbz:
I noticed these as I was cleaning up my untracked files (so that I could use `jj`).


chaychoong (approved):
Wow someone's spreading `jj` propaganda here

PR #84, opened 2022-11-14
personal-calls: add heroku-login recipe

glencbz:
I tried to deploy to staging and got reminded of the need to login. Add a make target (well, we use Make more as a recipebook I guess) that logs into heroku on the CLI and in Docker.

Also, I found a bug in the deploy-heroku target and fixed it.


chaychoong:
Good catch! Should we also make the login target a dep of the deploy target?

glencbz:
I don't think that's necessary, because you might already be logged in (e.g. you're deploying in a tight debugging loop), in which case that will be incredibly annoying. We _could_ try to skip it using actual targets. I doubt that's worth the effort, but if that's something you want to do, go for it

halfwhole (approved):
LGTM! have honestly not tested this out (don't have the permissions in any case), but the changes make sense

PR #86, opened 2022-11-30
Add theming to super-app

glencbz:
This copies the theming from `personal-calls/` so that it can be used in `super-app/`. We'll need this before we actually start working on the UI itself. There might be style issues because I haven't actually set up code quality tools on `super-app/` yet, but those should be hopefully minor and easy to fix later.

For reviewability, this is based off https://github.com/bettersg/call-home/pull/85, but we actually want to merge this into `master`.


apps/super-app/Makefile
```diff
@@ -0,0 +1,21 @@
...
+	cp -r ../../personal-calls/frontend/src/common src/
+# Do some simple transformations on the files
+	find src/common -name '*.[jt]s' | xargs -I% src/common/transform.sh %
+
+##@ Deployment
+
+##@ Helpers
+
+help: ## Display this help
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m\033[0m\n"} /^[a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
```
halfwhole, line 21: This command looks magical to me, it's pretty cool! Am not seeing `init` and `dev` in there though, do they require comments before they can show up in `help`?

Screenshot 2022-12-09 at 7 19 01 PM

apps/super-app/Makefile
```diff
@@ -0,0 +1,21 @@
...
+	npm ci
+
+dev:
+	npm start
+
+copy-common: ## Copy common files from the personal-calls common directory
+	rm -rf src/common
+	cp -r ../../personal-calls/frontend/src/common src/
+# Do some simple transformations on the files
+	find src/common -name '*.[jt]s' | xargs -I% src/common/transform.sh %
```
halfwhole, line 14: Is `transform.sh` available somewhere? I can't find it in the given directory in `personal-calls`

apps/super-app/src/common/contexts/ThemeProvider.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import { ThemeProvider as MuThemeProvider } from '@mui/material/styles';
```
halfwhole, line 2: nit: have a feeling this is intended to be `MuiThemeProvider` instead?

apps/super-app/package.json
```diff
@@ -3,6 +3,9 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@emotion/react": "^11.10.5",
+    "@emotion/styled": "^11.10.5",
```
halfwhole, line 7: Will these two deps be used? Don't appear to be imported anywhere

halfwhole (approved):
LGTM! Have a question about the mui v4/v5 split though, is the plan eventually to migrate `personal-calls/` over to v5?

apps/super-app/Makefile
glencbz: Gah, that's out of date. Will fix.

apps/super-app/Makefile
glencbz: Yeah it's sorcery to me too. It's frankly unreadable to me, so I'm just assuming it requires comments. I'll add them.

apps/super-app/package.json
glencbz: They're peer dependencies of @mui/material, see https://mui.com/material-ui/getting-started/installation/

apps/super-app/src/common/contexts/ThemeProvider.js
glencbz: Oh you're right.

glencbz:
> LGTM! Have a question about the mui v4/v5 split though, is the plan to eventually migrate personal-calls/ over to v5?

Yes, but that's a big headache because that requires running the codemods and making sure things still compile.
We'll cross that bridge when we get there.

PR #87, opened 2022-12-02
build: add terraform infra for staging

halfwhole:
This PR adds the terraform infra setups for the staging environment and some basic deploy scripts.

Run `make init` to initialise terraform using an existing remote state, and `make apply` to apply terraform configs.

- For simplicity, here we set up the GCP project and GCS bucket for terraform state beforehand, to be managed outside of terraform. For staging, I've already manually created the `call-home-staging` project (and given owner permissions to glenchoo@better.sg), as well as `call-home-staging-terraform-state-bucket`
- I've already tried applying the terraform configs to staging, and everything appears to work fine
- Have added cloud build service account IAM bindings to terraform, so that app engine deployments can work
- Staging and production are to be separated by a different file structure (`staging/` vs `production/`). Is this fine? I've looked a little into the alternatives and they seem a bit too gnarly, but if there's a better way I'm happy to try it!


glencbz:
> Staging and production are to be separated by a different file structure (staging/ vs production/). Is this fine? I've looked a little into the alternatives and they seem a bit too gnarly, but if there's a better way I'm happy to try it!

Yes, AFAIK this is the easiest and least error-prone way of doing it. I think I've mentioned [Terrform workspaces](https://developer.hashicorp.com/terraform/cloud-docs/workspaces) in the past, but it's relatively easy to invoke things in the wrong workspace and screw things up. @chaychoong has looked into them more deeply and his opinion is probably more informed than mine.

infra/staging/main.tf
```diff
@@ -0,0 +1,137 @@
...
+  settings {
+    tier = "db-f1-micro"
+    maintenance_window {
+      day = 7
+      hour = 20 # Sunday 8pm (UTC) = Monday 4am (UTC+8)
+    }
+  }
+}
+
+resource "google_sql_database" "strapi_db" {
```
glencbz, line 126: Wow, Terraform names. TIL `google_sql_database_instance` is the thing that costs compute resources and `google_sql_database` is an SQL database INSIDE the database_instance. How clear /s.

I assume we'll configure this database name in Strapi?

glencbz:
The approach of having variables + resources in a single file works for now, but we'll want to make this more DRY once we roll this out to production. In a follow up PR, let's refactor this to make it easier to do the Staging-Production split. My Terraform is a bit rusty, but IIRC the easiest way to do this would be:

- Create a single module that contains all of the data and resources and accepts variables.
- Having `staging/` and `production/` be directories with just a `variables.tfvars` and a `main.tf` that imports the module.

But if you come across something better, do go ahead.

infra/Makefile
```diff
@@ -0,0 +1,7 @@
+ENVIRONMENT?=staging
+
```
glencbz, line 2: When I tried to `make init`, I noticed my `gcloud` application default credentials expired lol. Let's add another recipe for setting that up.

```
gcloud-login:
	cd $(ENVIRONMENT)
	gcloud auth application-default login
```

infra/Makefile
```diff
@@ -0,0 +1,7 @@
...
+init:
+	cd $(ENVIRONMENT) && terraform init
```
glencbz, line 4: I initially thought this `&&` was unnecessary since `make` fails on the first non-zero exit, but unfortunately it seems like `make` executes each statement in its own subshell, so `cd` doesn't propagate the changed directory to subsequent statements. Yuck.

This `&&` is probably the cleanest way to do it then. I think a more typical pattern is to not use `cd` and to just invoke a Makefile at the root of the directory (either by cd-ing or by `make -C`), but this is a PITA for the staging + production setup (especially if we want to keep it DRY. Symlinking `Makefile` does work, but I think it's not worth the hassle).

infra/Makefile
```diff
@@ -0,0 +1,7 @@
...
+
```
glencbz, line 5: Suggestion: Add `&& cp secret.tfvars.template secret.tfvars`, which already gives us a file we can mess around with :)

infra/Makefile
```diff
@@ -0,0 +1,7 @@
...
+apply:
```
glencbz, line 6: Suggestion: rename to `tf-apply`, since "apply" is a terraform-specific thing

glencbz (changes requested):
I've tested this out by running `make init; make apply` and it looks like it works. Nice!

The only blocking comments are the ones asking for more docs/clarification. It's not clear to me what some of these resources are used for. It's pretty easy for cloud resources to go stale without us noticing, so I don't think it's possible to have _too much_ context.

I've left a few suggestions, you can choose to take or leave them.

glencbz:
If we could travel back in time, I would like to use the GCP resources in the same PR we introduce them, which would make it easier to see what each resource is used for. e.g. we could break some of this up into ~3 PRs + use the resources in the same PR:

- Setup Terraform file layout + create app engine app
- Add Strapi-specific things like the db + add scripts that deploy Strapi and verify it works
- Add App Engine deployment from GCB

But I don't think this is worth the churn of rewriting the PR. I think the PR is already really close to mergeable, so let's just add comments.

infra/Makefile
chaychoong: Yeap I just use cd xx && everywhere at work

chaychoong:
IMO repos should

> > Staging and production are to be separated by a different file structure (staging/ vs production/). Is this fine? I've looked a little into the alternatives and they seem a bit too gnarly, but if there's a better way I'm happy to try it!
> 
> Yes, AFAIK this is the easiest and least error-prone way of doing it. I think I've mentioned [Terrform workspaces](https://developer.hashicorp.com/terraform/cloud-docs/workspaces) in the past, but it's relatively easy to invoke things in the wrong workspace and screw things up. @chaychoong has looked into them more deeply and his opinion is probably more informed than mine.

Terraform workspaces is indeed the best way to go about this. IMO repository should be as environment agnostic as possible, so it's trivial to create/delete/reconfigure environments as necessary.

We can avoid making environment switching mistakes if we use make to specify the TF_WORKSPACE environment for each terraform command, as per https://developer.hashicorp.com/terraform/cli/config/environment-variables#tf_workspace

One possible way to do it is something like this:
```
ENVIRONMENT?=dev

TERRAFORM=TF_WORKSPACE=$(ENVIRONMENT) terraform
```

infra/staging/main.tf
```diff
@@ -0,0 +1,137 @@
...
+
+######################
+## Project Settings ##
+######################
+
+data "google_project" "project" {
+  project_id = var.project_id
+}
+
+// Enable Cloud SQL Admin API to connect to database
```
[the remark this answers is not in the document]
halfwhole: To be honest I'm not sure, in fact I'm not even sure if we need to enable this cloud SQL API at all? For some reason I think I needed it when doing the experiments, but I just tried terraform applying the DB infra with this API disabled and that worked well. Maybe it's needed for the GAE app to connect to the DB then? I can't be sure though until we actually deploy the app, so I've commented it out and left a TODO there - would this work for now?

infra/staging/main.tf
```diff
@@ -0,0 +1,137 @@
...
+  project = data.google_project.project.project_id
+  service = "cloudbuild.googleapis.com"
+}
+
+// Enable App Engine permissions for Cloud Build service account
+resource "google_project_iam_binding" "app_engine_admin" {
+  project = data.google_project.project.project_id
+  role = "roles/appengine.appAdmin"
+  members = [
+    "serviceAccount:${data.google_project.project.number}@cloudbuild.gserviceaccount.com"
```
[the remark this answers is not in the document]
halfwhole: Tried doing this but it failed because `data` couldn't be used within a variable :(

infra/staging/main.tf
```diff
@@ -0,0 +1,137 @@
...
+  service = "sqladmin.googleapis.com"
+}
+
+// Enable Cloud Build API to allow building for App Engine
+resource "google_project_service" "cloud_build_api" {
+  project = data.google_project.project.project_id
+  service = "cloudbuild.googleapis.com"
+}
+
+// Enable App Engine permissions for Cloud Build service account
```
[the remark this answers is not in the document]
halfwhole: yep, have added more descriptions for this in https://github.com/bettersg/call-home/pull/87/commits/d4f698af5e0e6f8d46d3b1396f5f44d76fec8e87

infra/staging/main.tf
```diff
@@ -0,0 +1,137 @@
...
+  ]
+  depends_on = [
+    google_project_service.cloud_build_api
+  ]
+}
+
+// Enable Service Account user permissions for Cloud Build service account
+resource "google_project_iam_binding" "service_account_user" {
+  project = data.google_project.project.project_id
+  role = "roles/iam.serviceAccountUser"
```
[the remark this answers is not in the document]
halfwhole: similar to above, added more descriptions in https://github.com/bettersg/call-home/pull/87/commits/d4f698af5e0e6f8d46d3b1396f5f44d76fec8e87

infra/staging/main.tf
```diff
@@ -0,0 +1,137 @@
...
+  depends_on = [
+    google_project_service.cloud_build_api
+  ]
+}
+
+#############
+## Storage ##
+#############
+
+resource "google_storage_bucket" "app_bucket" {
```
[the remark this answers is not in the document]
halfwhole: IIRC when running the experiments I had difficulties deploying to GAE using the default app bucket as it wasn't managed by terraform, so I created a separate app bucket for that. But at that time I think I was trying out quite a different terraform setup - now I seriously doubt that it's still needed :x

Have commented out this resource for now, perhaps we could revisit it after we try deploying strapi to GAE?

infra/staging/main.tf
halfwhole: yep, the database name will be configured for the Strapi app through env vars in [app.template.yaml](https://github.com/glencbz/call-home-cms-experiments/pull/6/files#diff-a63b4136b40f69b3e3523db48febd1c844cbbfe61bc68fd0e6d82fc80f7c31c8)!

infra/Makefile
halfwhole: https://github.com/bettersg/call-home/pull/87/commits/d082e99765c55115663117c99cc9703b30ad73e2

infra/Makefile
halfwhole: https://github.com/bettersg/call-home/pull/87/commits/d082e99765c55115663117c99cc9703b30ad73e2

infra/Makefile
halfwhole: https://github.com/bettersg/call-home/pull/87/commits/d082e99765c55115663117c99cc9703b30ad73e2

halfwhole:
Have commented out some probably-useless terraform infra, added some documentation for terraform resources, and made small adjustments to the Makefile. PTAL!

On the issue of staging/prod separation: if we use terraform workspaces, in which project's bucket should the remote state go? Initially I was thinking that there could be 2 projects (e.g. `call-home-staging`/`call-home-production`), and each of these could have their own state bucket; with workspaces though, it seems (though I'm really not sure!) that both workspaces for staging/prod must share a single backend, and so a single bucket under a single GCP project - how is a situation like this usually best handled?

halfwhole:
> Different workspaces share the same bucket/key, and locking table for state.
> 
> Don't create the state bucket with terraform. At least not in this project's terraform.

I understand that different workspaces share the same bucket, and the original intention was always to create the state bucket manually (as it already has been!).

What I'm more curious to know is about how to reconcile these two things:
1. The current plan is to create two GCP projects (manually), `call-home-staging` and `call-home-production`;
2. If we use terraform workspaces, both staging and production will share only one terraform state bucket.

This raises the question, under which of the two projects should this single manually-created state bucket go? I can think of some initial possibilities:
- Forcefully place this state bucket under either `call-home-staging` or `call-home-production` - but would this approach be clumsy and far too unclean?
- Merge `call-home-staging` and `call-home-production` into a single project `call-home`, thereafter to contain the single manually-created state bucket and all its terraform-managed infrastructure, be they for staging or production - but would this threaten the isolation of staging and production environments?
- Spin off a third manually-created project (e.d. `call-home-terraform-state`) to hold this terraform state bucket exclusively - but would this be a bit more overhead than we need to deal with?

That's what I was hoping to find out!

chaychoong:
Oh I see, I misunderstood the previous comment. There's usually a meta project (maybe call-home-infrastructure) that contains the tfstate and other common infra stuff, like secrets, shared assets, etc.

infra/staging/main.tf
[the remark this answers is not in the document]
glencbz: I think it's probably better to just delete the lines instead of commenting them out. These sorts of "maybe we need this idk" comments have a tendency to get stale, especially if

a) we don't consciously look at them (maybe doesn't apply, since you are looking at this actively)
b) we don't leave clues as to when we should have already deleted them (e.g. "We can delete this if Strapi runs fine in staging without it."), and by the time we see them again we've forgotten the original context.

In general, I prefer to have these sorts of comments in a more ephemeral medium like the PR message, design doc or ticket. But it's not a big enough deal to block submission.

infra/staging/main.tf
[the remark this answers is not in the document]
glencbz: Ah, yeah that would make sense. I guess I was thinking about using the service account as a data source (https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/service_account) and then referencing its `.email` property. Nice to have, but not a blocker.

glencbz (approved):
Thanks! LGTM, we can take the code changes as-is, though I have a preference for removing the comments.

I'd prefer not to take the 3 commits separately (since the later 2 are just review fixups), so either:

a) use squash and merge
b) squash the commits on your end and force push (don't listen to the naysayers, force pushing is fine), then use "rebase and merge". Bonus points if you add most of the PR description to the commit message (it's nicer to have commit descriptions on the commit than on the PR)

infra/staging/main.tf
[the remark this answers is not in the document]
glencbz: We may not revisit it at all, so let's just delete this :)

glencbz:
Thanks both. To summarise, here are the main things from the discussion above:

 ## Which GCP project should manage the Terraform state bucket?
@halfwhole had some suggestions in https://github.com/bettersg/call-home/pull/87#issuecomment-1344989913, @chaychoong recommends using a separate project here https://github.com/bettersg/call-home/pull/87#issuecomment-1345001668. Both agree that the bucket shouldn't be managed by this Terraform config, and possibly not by Terraform at all.

I'm keen on the separate project approach (maybe `call-home-infra`). This probably has the best separation of concerns, though it's more overhead than necessary. But AFAICT it's a pain to migrate GCS buckets between projects, so let's just do it the "right" way the first time.

I'm okay for `call-home-infra` to not be managed by Terraform, I don't think it will matter much in practice.

## How to separate staging and production

@glencbz suggests using separate directories in https://github.com/bettersg/call-home/pull/87#issuecomment-1336272260, @chaychoong prefers setting the `TF_WORKSPACE` variable https://github.com/bettersg/call-home/pull/87#issuecomment-1336317377. I think we both agree on keeping the config environment-agnostic, and I don't think there's a real difference between modules vs workspaces in that regard.

Here's what I see so far, let me know if I'm missing anything, especially if there are any particularly strong reasons to use one over the other.

**A good thing about workspaces:** having a single directory is more convenient for some tooling, e.g. we have a single `Makefile` at the root of the tf config instead of having 1-3 `Makefile`s to manage 2 sets of tf configs.

**Things that I don't really like about workspaces, but can be worked around pretty easily:**
- It's relatively easy to forget to set the workspace or run in the wrong workspace as opposed to the the wrong directory. This is mostly obviated by setting the right things in `make` and not running `terraform` manually.
- It moves state out of Terraform config, which makes it harder to see what's under Terraform control. With a shared module, every environment is its own directory, so simply inspecting the directory structure tells you what environments you have, whereas Terraform workspaces require some additional setup. We could manage the setup and such with `make` though. We're already leveraging `make` for automation and standardisation, so I think this is okay.

**A bad thing about workspaces**
Workspaces reuse all of the config, so it requires each environment to be nearly-identical to each other; minor differences can be handled by checking `terraform.workspace` in the config (https://developer.hashicorp.com/terraform/language/state/workspaces#current-workspace-interpolation).

This makes me feel very uneasy, because this makes it harder to express significant differences between environments.
A possible example is that we want to expose production to end users, so it needs a DNS + load balancing setup (or similar), but staging doesn't need that.

(I say _possible_, since DNS isn't particularly high stakes; it's not the end of the world if we give staging a DNS setup, and arguably we _should_ have one since it makes DNS changes previewable. I'm just giving an example for argument's sake.)

With modules, this is really easy to express:

```
.
├── modules/
│   ├── strapi
│   └── dns
└── environments/
    ├── staging (only use the strapi module)
    └── production (use both modules)
```

What's the equivalent in workspaces?
- One config per "module" and apply multiple configs to a GCP project? Sounds much harder to manage than one config per project and has more opportunity for configs to step on each other's toes
-  One config and conditionally enable all of the DNS resources based on `terraform.workspace`? Sounds extremely messy.
- One config, have one module for all DNS resources, then enable the module using `terraform.workspace`? Sounds clean, but if we were doing this, why wouldn't we just use modules and separate directories?

I find it hard not to see workspaces as locking us into a less flexible setup than modules + separate directories. Will we need the flexibility? Maybe not, but it's nice to have the option.

infra/staging/main.tf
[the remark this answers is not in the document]
halfwhole: Have removed the commented-out cloud SQL API in https://github.com/bettersg/call-home/pull/87/commits/a5d163e756a4429ef771f8ce2174d2249c6f2714

infra/staging/main.tf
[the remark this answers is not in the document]
halfwhole: Ah I didn't realise that the service account data source existed! I've added this data source into https://github.com/bettersg/call-home/pull/87/commits/a5d163e756a4429ef771f8ce2174d2249c6f2714 and then tried reapplying terraform configs, and terraform tells me that there are no changes found, so the refactoring looks good

infra/staging/main.tf
```diff
@@ -0,0 +1,112 @@
...
+// Enable Cloud Build API and grant its service account App Engine and Service Account User
+// permissions to allow developers to build and deploy applications to App Engine using gcloud SDK.
+// See: https://cloud.google.com/build/docs/deploying-builds/deploy-appengine
+resource "google_project_service" "cloud_build_api" {
+  project = data.google_project.project.project_id
+  service = "cloudbuild.googleapis.com"
+}
+
+data "google_service_account" "cloud_build_service_account" {
+  account_id = google_project_service.cloud_build_api.id
```
glencbz, line 52: Oh interesting, I didn't know you could reference the cloud build account like this.

glencbz (approved):
Looks even better. Thanks!
My thoughts on how to integrate this are the same as before. Let's squash the changes (either manually or with the GitHub UI) before merging.

PR #89, opened 2022-12-05
super-app: add simple development scripts, show frontend-backend integration

glencbz:
- Add simple development scripts to `super-app-cms`.
- Add a simple resource to `super-app-cms` and query it from `super-app`

This depends on https://github.com/bettersg/call-home/pull/88 and https://github.com/bettersg/call-home/pull/85.


services/super-app-cms/Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+##@ Dev
+
+# NOTE: This affects where the app runs, but NOT where its own routes point.
+# That is configured in config/server.js, which gets baked in at build-time, not
+# run-time. So to properly run on PORT in development, you have to run `build`
+# before `dev`, e.g. `make PORT=4000 build && make PORT=4000 dev`.
+
+dev: ## Run Strapi with autoReload enabled
+	PORT=$(PORT) npm run develop
```
halfwhole, line 17: would another `init` command that runs `npm ci` and copies `.env.example` to `.env` be helpful here?

apps/super-app/src/services/CmsSmokeTest.ts
```diff
@@ -0,0 +1,30 @@
...
+    attributes: T;
+  };
+}
+
+interface SmokeTestView {
+  test: boolean;
+}
+
+async function isUp(): Promise<boolean> {
+  const response = await fetch(smokeTestEndpoint);
```
halfwhole, line 17: maybe it'd be nice to have a way to handle an error here, if the strapi API endpoint is down? i'm getting response status 500 with the error message `Proxy error: Could not proxy request /api/smoke-test from localhost:3000 to http://localhost:1337 (ECONNREFUSED)` locally when strapi dev is not up

but then again this could be good enough. i'm also not familiar with smoke tests and don't know how important it is for this code to be robust, is this test meant to be permanent?

services/super-app-cms/README.md
```diff
@@ -1,57 +1,3 @@
...
-## ✨ Community
-
-- [Discord](https://discord.strapi.io) - Come chat with the Strapi community including the core team.
-- [Forum](https://forum.strapi.io/) - Place to discuss, ask questions and find answers, show your Strapi project and get feedback or just talk with other Community members.
-- [Awesome Strapi](https://github.com/strapi/awesome-strapi) - A curated list of awesome things related to Strapi.
-
----
-
-<sub>🤫 Psst! [Strapi is hiring](https://strapi.io/careers).</sub>
+[Strapi](https://docs.strapi.io)-powered content management!
```
halfwhole, line 3: It could be nice to document how to configure strapi permissions here eventually - when setting up the smoke test, I was getting 403 errors because I forgot to allow the 'find' permission for 'smoke-test' on the strapi admin dashboard

halfwhole (approved):
Just tried running it locally, it all works well!

services/super-app-cms/Makefile
glencbz: That's a good idea, I'll add that.

services/super-app-cms/README.md
glencbz: Yeah, seems like a good thing to have in the docs themselves instead of just in the commit message. I'll do that.

apps/super-app/src/services/CmsSmokeTest.ts
glencbz: This is meant to be a rudimentary simulation of making a call to Strapi to get content (looks like I need to document that better). I'm not sure if there's a good way to handle errors at this level; we _want_ to report that something went wrong so that we can surface it on the UI. We could make this never throw, but for better or worse (probably worse, I hate exception throwing as a pattern), throwing exceptions is an widespread pattern in JS (e.g. it has first class support in React and allows us to display an "oops something went wrong" if we want, like in [`personal-calls/`](https://github.com/bettersg/call-home/blob/5fdc009020fb058da2c40d9f6f0bcffae5af1a11/personal-calls/frontend/src/App.tsx).

I should document that this isn't meant to be permanent; I don't foresee a smoke test like this being useful unless we also add additional telemetry/error reporting e.g. Sentry.

PR #98, opened 2022-12-12
Fix super-app CI

glencbz:
You might be asking "fix? we had CI?". Well, I committed the CI stuff right to `master` (the GitHub UI told me it was fine!!), but I don't know the first thing about GitHub actions and got it wrong. So now I need to file this PR and admit my hubris...


halfwhole (approved):
lgtm!

PR #101, opened 2022-12-14
Add simple container element

glencbz:
This copies the `Container` element from `personal-calls/` to `super-app/`. The container is what's responsible for the basic layout of a Call Home scene, namely:

- Horizontal centering
- Stacking elements vertically
- Using 100% of the viewport height at minimum


glencbz:
Superseded by https://github.com/bettersg/call-home/pull/102

PR #102, opened 2022-12-14
Super app: add container element and client-side routing

glencbz:
This adds some more scaffolding to the super app, specifically:

- Copy the `Container` element from `personal-calls/` to `super-app/`. The container is what's responsible for the basic layout of a Call Home scene, namely:
  - Horizontal centering
  - Stacking elements vertically
  - Using 100% of the viewport height at minimum
- Add a simple react-router setup


halfwhole (approved):
lgtm!

PR #103, opened 2022-12-18
created footer component

### LINKED ISSUE #99: [Super app] Create footer component
Implement the navigation footer component as illustrated in the picture below/[mockup](https://www.figma.com/file/aAmKbEWb314Keeki3xlcfd/Call-Home-(MVP)?node-id=26673%3A76315). In particular:

- Width should be 100% of parent width
- Height should be fixed
- The footer should have 3 clickable icons equally spaced apart
- When an icon is clicked, it should become highlighted and other icons should become un-highlighted

Icons should be exportable from Figma, approach UXers for help if necessary.

Image

e056:
Fixes #99 

created footer component


apps/super-app/src/App.tsx
```diff
@@ -1,17 +1,28 @@
...
     <ThemeProvider>
       <PrimaryButton>Primary</PrimaryButton>
       <NeutralButton>Neutral</NeutralButton>
       <ErrorButton>Error</ErrorButton>
+      <NavBar/>
+      <Routes>
+          <Route path="/callhome" element={<CallHome />}></Route>
+          <Route path="/support" element={<Support />}></Route>
+          <Route path="/more" element={<More />}></Route>
+        </Routes>
```
glencbz, line 25: Oh neat, this actually uses the React router state to determine the active state. I originally envisioned that the nav bar would just use internal state, but this is okay too (TIL about NavLink, good find!). I didn't intend for you to add React router because this would conflict quite drastically with https://github.com/bettersg/call-home/pull/102, which also adds React router. I _think_ there's no need to change the structure of this PR, but FYI I will tweak this  and https://github.com/bettersg/call-home/pull/102 to be more consistent.

apps/super-app/src/common/components/CallHome.tsx
```diff
@@ -0,0 +1,5 @@
+const CallHome = () => {
+  return <h1>Call Home Page placeholder</h1>;
+};
```
glencbz, line 3: For consistency, prefer using named functions when declaring React components, e.g.

```
function CallHome() {
}
```

This used to matter a lot because React couldn't read the component name from an anonymous function like this. so this is the style I use (though I think this isn't the case any more). I also find that a bit easier to read than the fat arrow syntax, and at any rate let's not mix styles if we don't have to.

apps/super-app/src/common/components/navbar/NavBar.css
```diff
@@ -0,0 +1,44 @@
+@import url("https://fonts.googleapis.com/css?family=Hind");
+a:visited {
+  text-decoration: none;
+}
```
glencbz, line 4: Let's apply this to all `<a>` elements, i.e.
```
a {
  text-decoration: none;
}
```

apps/super-app/src/common/components/navbar/NavBar.css
```diff
@@ -0,0 +1,44 @@
...
+
+.navbar {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0px 32px 10px;
+  gap: 56px;
+  position: absolute;
```
glencbz, line 13: This won't matter now, but let's drop the absolute positioning. It won't work well when we're positioning multiple elements on the page and I don't think it makes a difference to the demo.

apps/super-app/src/common/components/navbar/NavBar.css
```diff
@@ -0,0 +1,44 @@
...
+a:visited {
+  text-decoration: none;
+}
+
+.navbar {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0px 32px 10px;
```
glencbz, line 11: I suggest using `rem` as the unit instead of `px`. `rem` scales a bit better do different screen sizes.

apps/super-app/src/common/components/navbar/NavIcon.tsx
```diff
@@ -0,0 +1,17 @@
+import { NavLink } from "react-router-dom";
+function NavIcon(props:any) {
+  return (
+    <div>
+      <NavLink
+        to={props.link}
+        className={({ isActive }) =>
+          isActive ? "navbar-icon-active" : "navbar-icon"
```
glencbz, line 8: Instead of switching between two classes, let's switch between one class and drop the function like so:

```
className={`navbar-icon${isActive ? ' navbar-icon-active' : ''}}
```

This will simplify the CSS too (see that comment for more explanation).

apps/super-app/src/common/components/navbar/NavBar.css
```diff
@@ -0,0 +1,44 @@
...
+}
+
+.navbar-icon-text {
+  display: block;
+  font-size: 16px;
+  font-weight: bold;
+  font-family: "hind";
+}
+
+.navbar-icon-active {
```
glencbz, line 37: This ends up repeating most of the styles from `.navbar-icon`, which isn't really what we want.
Instead, we can simplify this by thinking of classes as a "is-a" relationship, e.g. an active navbar icon is both active and a navbar icon, so we can make it two classes, like:


```
.navbar-icon {
  display: inline-block;
  color: #bec7ca;
  text-decoration: none;
  text-align: center;
  align-items: center;
  line-height: 1rem;
}

.navbar-icon.navbar-icon-active {
  color: #297b94;
}
```

The higher specificity will ensure that the `-active` variant always takes precedence.

apps/super-app/src/common/components/navbar/NavIcon.tsx
```diff
@@ -0,0 +1,17 @@
+import { NavLink } from "react-router-dom";
+function NavIcon(props:any) {
```
glencbz, line 2: Please use a more suitable type, [personal-calls/frontend/src/common/components/Container.tsx](https://github.com/bettersg/call-home/pull/102/files#diff-931083105ec009d12e6eaa8b9295fbac95aeb33cdcde68414079b8af1bbfafa8) has an example. You can also rely on the CI check to warn on this https://github.com/bettersg/call-home/actions/runs/3724490539/jobs/6316677421.

apps/super-app/src/index.tsx
```diff
@@ -3,14 +3,15 @@ import ReactDOM from 'react-dom/client';
...
 
 const root = ReactDOM.createRoot(
   document.getElementById('root') as HTMLElement
 );
 root.render(
-  <React.StrictMode>
+  <Router>
     <App />
-  </React.StrictMode>
+  </Router>
```
glencbz, line 14: Why does this remove `<React.StrictMode>`? I would expect `<Router>` to be inside `<React.StrictMode>`, not that we would remove it altogether.

apps/super-app/src/common/components/navbar/NavIcon.tsx
```diff
@@ -0,0 +1,17 @@
...
+          isActive ? "navbar-icon-active" : "navbar-icon"
+        }
+      >
+        {props.icon}
+        <span className="navbar-icon-text">{props.text}</span>
+      </NavLink>
+    </div>
+  );
+}
+export default NavIcon;
```
glencbz, line 17: I suggest putting this in the same file as Navbar. We're never going to use this outside of Navbar anyway :)

apps/super-app/src/common/components/More.tsx
```diff
@@ -0,0 +1,6 @@
+const More = () => {
+    return <h1>More Page placeholder</h1>;
+  };
+  
+  export default More;
+
```
glencbz, line 6: There are a few formatting errors like this. It's pretty easy to fix (just use `make format`). You can check the CI for more info https://github.com/bettersg/call-home/pull/103/checks.

apps/super-app/src/common/components/navbar/NavBar.css
glencbz: Also, do we need the bottom padding? The mockup looks like the icons are vertically centered.

apps/super-app/src/common/components/navbar/NavBar.css
```diff
@@ -0,0 +1,44 @@
...
+  position: absolute;
+  width: 360px;
+  height: 72px;
+  left: 16px;
+  top: 104px;
+  background-color: #f3fafc;
+}
+
+.navbar-icon {
+  display: inline-block;
```
glencbz, line 22: Let's use `display: flex` instead; it's easier to control the formatting than by manipulating line heights in inline-block elements :)

apps/super-app/src/index.tsx
```diff
@@ -3,14 +3,15 @@ import ReactDOM from 'react-dom/client';
...
 import App from './App';
 import reportWebVitals from './reportWebVitals';
+import { BrowserRouter as Router } from 'react-router-dom';
 
 const root = ReactDOM.createRoot(
   document.getElementById('root') as HTMLElement
 );
 root.render(
-  <React.StrictMode>
+  <Router>
```
glencbz, line 12: This is my personal taste, but let's move this into the `App` component instead of outside of it. I would prefer to be able to see all of the providers at once instead of having some providers inside `App` and some outside of it.

glencbz (changes requested):
Great first PR. Thanks for the effort :) I was surprised to see React router; this will require some more tweaking with regards to the rest of the routing, but this shows how we could implement this using React router, which is great.

Re: the commit message, please use the imperative mood for consistency; that's what most https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/ folks use.

There are two options for making changes to this PR:
- (Beginner friendly option) Add your fixes in another commit. If you do this, I would strongly prefer we use "Squash and merge" to merge this into `master`. If we do that, the commit messages will be lost, so do leave the "real" commit message in the PR description. I think GitHub knows how to populate the final commit message with that.
- (More advanced, aka more Git learning) Rewrite this commit with the new message and changes. I like this because it's cleaner and better matches what we actually want to see in the final history. However, if you're not comfortable rewriting Git history and would prefer not to do this now, that's okay.

glencbz:
The react-router bits of this PR conflict quite heavily with https://github.com/bettersg/call-home/pull/102, and are also responsible for the CI failures (e.g. https://github.com/bettersg/call-home/actions/runs/3728660563/jobs/6323865978). I think these will be pretty difficult to resolve, especially because I haven't 100% figured out what we should be doing with react-router or the test suite.

On the other hand, the footer component looks pretty close to being mergeable 🎉  and I think that's useful on its own. Here's what I suggest: let's focus on just getting the footer component cleaned up. After merging, I'll figure out what to do about react-router on my end. Does that sound good?

apps/super-app/src/components/navbar/NavBar.css
```diff
@@ -0,0 +1,38 @@
...
+  text-decoration: none;
+}
+
+.navbar {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0px 32px 10px;
+  gap: 56px;
```
glencbz, line 12: The left and right alignment looks off with the `gap`. I think we can remove it.

apps/super-app/src/components/navbar/NavBar.css
```diff
@@ -0,0 +1,38 @@
...
+a {
+  text-decoration: none;
+}
+
+.navbar {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0px 32px 10px;
```
glencbz, line 11: What's the reasoning behind these values? It doesn't seem necessary for centering.

apps/super-app/src/components/navbar/Navbar.tsx
```diff
@@ -0,0 +1,31 @@
+import './NavBar.css';
+import { NavLink } from 'react-router-dom';
+import { ReactComponent as CallHomeIcon } from '../img/callhomeicon.svg';
+import { ReactComponent as SupportIcon } from '../img/supporticon.svg';
+import { ReactComponent as MoreIcon } from '../img/moreicon.svg';
+
+export const NavBar = () => {
```
glencbz, line 7: I think this was also meant to be converted to `function`?

apps/super-app/src/index.tsx
```diff
@@ -3,17 +3,17 @@ import ReactDOM from 'react-dom/client';
...
 
 const root = ReactDOM.createRoot(
   document.getElementById('root') as HTMLElement
 );
 root.render(
   <React.StrictMode>
-    <App />
+    <BrowserRouter>
+      <App />
+    </BrowserRouter>
```
glencbz, line 15: After experimenting a bit on my end, I think having `BrowserRouter` wrapping `App` looks fine.

glencbz (changes requested):
This looks quite good :) The React component itself looks mergeable; I think there are just some minor CSS issues, but otherwise this is mergeable.

apps/super-app/src/components/navbar/NavBar.css
e056: Hi Glen! 
The padding is so that there is a spacing between the icons at the edges and the sides of the box (it will look like the pic i attached below). However, I understand that using px for the padding is not scalable, I was wondering if there is another way to design this? Thanks!
image

glencbz: Ah I meant vertical centering. The horizontal padding is fine; I don't understand the discrepancy between top and bottom padding.

The simple way to avoid px is to use `rem` :) you can eyeball the values and it should be fine. Don't feel the need to replace px with other units at this time though, we can adjust them later.

glencbz (approved):

PR #114, opened 2023-01-12
build: enable cloud sql admin api

halfwhole:
The Cloud SQL Admin API must be enabled for the App Engine app to connect to a Cloud SQL instance, see https://cloud.google.com/sql/docs/mysql/connect-app-engine-standard

This follows up on https://github.com/bettersg/call-home/pull/87#discussion_r1038876358, apparently the API needed to be enabled after all! After applying the new terraform config in this PR, the 500 error on strapi staging goes away. :)

### DIFF
--- a/infra/staging/main.tf
+++ b/infra/staging/main.tf
@@ -40,6 +40,13 @@ data "google_project" "project" {
   project_id = var.project_id
 }
 
+// Enable Cloud SQL Admin API to allow App Engine to connect to the database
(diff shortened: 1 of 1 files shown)


glencbz (approved):
Interesting! I see that the logs say that the Cloud SQL Admin API must be enabled, so this does make sense.

I frankly have no idea _why_ we need the Admin API though; https://cloud.google.com/sql/docs/mysql/connect-app-engine-standard doesn't explain it, but I wouldn't be surprised if there's just some bad design somewhere.

PR #116, opened 2023-01-31
Add more to support detail page

glencbz:
- Add logo
- Tweak fonts and alignment of the blurb header
- Add navigation header


apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -1,17 +1,28 @@
...
 import { PrimaryButton } from '../common/components/RoundedButton';
 import './SupportDetailWip.css';
 
-function Header() {
-  return <nav className="support-header">Navigation</nav>;
+function Header(props: { title: string }) {
+  const { title } = props;
+  return (
+    <nav className="support-header">
+      <ArrowBackIosNewIcon />
```
chaitanyabaranwal, line 22: Can we add some spacing between the arrow and the header? It looks a little cramped right now.

apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -1,17 +1,28 @@
...
+      <Typography variant="h4">{title}</Typography>
```
chaitanyabaranwal, line 23: Is this meant to have the same font type as the blurb title? I don't have any frontend sense but was just wondering if the header font should be different from the blurb.

apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -74,14 +85,14 @@ function DescriptionSection(props: DescriptionProps) {
   return (
     <main className="support-description-container">
       <section className="support-description-header">
-        <img src={logo} style={{ flexBasis: '8rem' }} />
+        <img src={logo} style={{ flexBasis: '6rem', width: '1px' }} />
```
chaitanyabaranwal, line 88: I am not sure why this changed (probably just better sizing?), but I don't see any issues with this change.

chaitanyabaranwal (approved):
Added some minor comments but looks good to me otherwise. Really don't have a frontend sense so another set of eyes might be good (or we can always update things later).

apps/super-app/src/routes/SupportDetailWip.tsx
stevanussatria: That's fine for now, we can fine tune it later. We can cross check against the mocks

stevanussatria (approved):
Aside from wanting to be sure that we're not strict with "no null values", looks good!

apps/super-app/src/routes/SupportDetailWip.tsx
glencbz: We'll tweak these in the next round of changes :)

apps/super-app/src/routes/SupportDetailWip.tsx
glencbz: Ditto, we'll tweak these in the next round of changes :)

apps/super-app/src/routes/SupportDetailWip.tsx
glencbz: Yeah, it seemed like better spacing as I was eyeballing it. The `width: 1px` is there to give an initial width that the flex box can grow later.

glencbz:
> Aside from wanting to be sure that we're not strict with "no null values"

@stevanussatria I'm not sure what you're referencing here, though I take it that the PR looks okay to you?

stevahnes:
> > Aside from wanting to be sure that we're not strict with "no null values"
> 
> @stevanussatria I'm not sure what you're referencing here, though I take it that the PR looks okay to you?

yeap PR looks good! Just that a colleague of mine once tried to never have any `null` values haha, so wondering if we are aiming for that level of strictness in the code.

stevahnes (approved):

glencbz:
> a colleague of mine once tried to never have any null values haha, so wondering if we are aiming for that level of strictness in the code.

Yes, we are. If this is a general question, it might be better to ask this on Slack, or if this is a specific thing, it would be better to anchor the comment to a bit of code.

PR #119, opened 2023-02-11
Update infra commands to serve frontend on GCS

### LINKED ISSUE #133: [Super app] Create build and deploy process for frontend
Figure out how to build the frontend and serve it from GCP.
The "best way" is probably to use GCS (because that doesn't require any compute), but let's figure out what the "best way" is and PoC that.

Add a script to `apps/super-app/Makefile` that builds and deploys the frontend to staging.

A single JavaScript bundle is fine, aka code splitting the frontend is out of scope. We'll figure that out in the future (if we want it at all).

chaitanyabaranwal:
Fixes #133 

This PR makes a few changes:
1. It updates the staging Terraform file to add new commands for creating a GCS bucket. This bucket will be used to store frontend assets.
2. It adds a new `Makefile` in the `apps/` directory for building the frontend.
3. It adds an `.env.staging` file to store the Frontend bucket name and pass it down to Terraform. 
4. It adds a top-level `Makefile` to deploy the frontend assets into the GCS bucket.

The PR does not make changes to the react-router configuration. As a result, the deployed assets do not really work when we go the GCS link. This will be addresses in a follow-up PR.

Some things that I want to discuss:
1. Currently, we need to run the `source .env.staging` command to get the bucket name, and only then can we deploy to Terraform. I've added that in `infra/README/md` but wondering if there's a better way to throw variables around directories.


infra/Makefile
```diff
@@ -8,3 +12,6 @@ init:
 
 tf-apply:
 	cd $(ENVIRONMENT) && terraform apply -var-file="secret.tfvars"
+	
+frontend-deploy:
+	cd $(FRONTEND_SRC) && npm run build && gsutil cp -r $(FRONTEND_BUILD_FOLDER)/* gs://$(TF_VAR_frontend_bucket_name)
```
glencbz, line 17: It feels like the infra "module" is overstepping its bounds by going to the frontend module and running its scripts. It would be a bit odd if a frontend dev has to run scripts from `infra/` to deploy a new version, especially since deployment isn't typically an infra concern.

Maybe this suggests that we should put this script in a top-level Makefile that can see all the other modules, or this script should be in the frontend Makefile (but then how would it see the bucket from terraform? Hmm).

WDYT?

infra/staging/main.tf
```diff
@@ -119,3 +124,39 @@ resource "google_sql_user" "user" {
...
+}
+
+# Add proper permissions to bucket
+resource "google_storage_bucket_iam_member" "viewer" {
+  bucket = google_storage_bucket.call_home_frontend.name
+  role   = "roles/storage.objectViewer"
+  member = "allUsers"
+}
+
+resource "google_storage_bucket_iam_member" "admin-iam" {
```
glencbz, line 152: Do use a more specific name for this, in the use_case_then_name style, like `frontend_bucket_viewer`. The reason for that is that terraform addresses are global (so the chance of conflict is high) and you don't always get to see the full details of the resource (so the name has to be self-descriptive to aid understanding).

glencbz (changes requested):

infra/Makefile
chaitanyabaranwal: Would it make sense to separate the frontend specific commands into the frontend Makefile? That would bring in a dependency (the infra commands need to be executed before the frontend make commands), but I think it separates concerns well.

apps/super-app/package.json
```diff
@@ -53,5 +53,6 @@
     "eslint": "^7.32.0",
     "eslint-config-prettier": "^8.5.0",
     "prettier": "^2.8.1"
-  }
+  },
+  "homepage": "."
```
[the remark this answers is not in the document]
chaitanyabaranwal: Will keep this in mind next time!

infra/Makefile
```diff
@@ -1,5 +1,9 @@
 ENVIRONMENT?=staging
 
+FRONTEND_SRC?=../apps/super-app
+FRONTEND_BUILD_FOLDER?=build
+TF_VAR_frontend_bucket_name?=call-home-frontend
```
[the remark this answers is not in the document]
chaitanyabaranwal: It's a little hard to obtain the `terraform output` results in a different directory (the frontend directory). How should I go about doing that? I tried exporting the `terraform output` results into an environment variable but that doesn't seem to work.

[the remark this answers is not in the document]
glencbz: > It's a little hard to obtain the terraform output results in a different directory (the frontend directory)

Why would we need to use the result of `terraform output` in the frontend directory?

As an aside, perhaps there is some Make trouble afoot. Each line in the `Makefile` is executed it its own subshell, so the environment variables aren't shared. Here's a recipe that worked for me:

```
	FOO="$$(cd staging && terraform output --raw frontend-bucket-name)"; \
	echo $$FOO
```

The `--raw` removes the extra quotes from `terraform output`. As an aside, these little issues are starting to make me a bit wary of writing scripts in Makefiles. Maybe it's not so bad yet, but we should be careful that it doesn't become a source of unintentional complexity.

infra/Makefile
glencbz: > Would it make sense to separate the frontend specific commands into the frontend Makefile?

Yes, I think so. This would be a much cleaner separation of concerns.

> That would bring in a dependency (the infra commands need to be executed before the frontend make commands)

I'm not sure I follow this line of thinking. What are you envisioning?

The idea in my head doesn't seems to have the dependency you're thinking of:
- `frontend` would have a script that builds and outputs to `frontend/build/` (basically `npm build`)
- `infra/` would have a script that creates the frontend bucket (i.e. by running `terraform apply`)
- the top level makefile would know how to get the frontend bucket name and use it to deploy a new version of the frontend

glencbz:
@chaychoong what do you think about how we should coordinate between the `frontend` and `infra` modules?

infra/staging/main.tf
```diff
@@ -119,3 +124,44 @@ resource "google_sql_user" "user" {
...
+resource "google_storage_bucket_iam_member" "frontend_bucket_viewer" {
+  bucket = google_storage_bucket.call_home_frontend.name
+  role   = "roles/storage.objectViewer"
+  member = "allUsers"
+}
+
+resource "google_storage_bucket_iam_member" "frontend_bucket_admin" {
+  bucket = google_storage_bucket.call_home_frontend.name
+  role   = "roles/storage.admin"
+  member = "projectOwner:${data.google_project.project.project_id}"
```
glencbz, line 156: Why do we need this permission? I'd assume that project owners already have admin permissions over the GCS bucket.

infra/staging/main.tf
```diff
@@ -119,3 +124,44 @@ resource "google_sql_user" "user" {
...
+}
+
+#############
+## Outputs ##
+#############
+
+# Output the built frontend HTML index
+output "frontend-url" {
+  description = "Frontend URL"
+  value = "https://storage.googleapis.com/${google_storage_bucket.call_home_frontend.name}/index.html"
```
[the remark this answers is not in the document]
chaitanyabaranwal: Removed! I had earlier added it for convenience (I could just copy the output link to test the site).

infra/staging/main.tf
chaitanyabaranwal: Removed.

chaitanyabaranwal:
@glencbz Updated the PR description with the changes! Would need some discussion on whether the current way to pass the bucket name works.

glencbz:
Thanks, the new PR description is really helpful :)

glencbz (changes requested):
Looks good, we can merge this once we get some finishing touches

> Currently, we need to run the source .env.staging command to get the bucket name, and only then can we deploy to Terraform. I've added that in infra/README/md but wondering if there's a better way to throw variables around directories.

Yeah, this does seem a little bit error-prone. Not sure if it's bad enough that I'd drop the approach, but I'll need to try it myself.

The best alternative I can think of is to modify the makefile recipes to read the variables at runtime. That solves the problem of having to do 2 different steps. e.g. we could imagine something like:

```make
ENVIRONMENT?= staging

deploy-frontend:
  source ../../..env.${ENVIRONMENT}; gsutil bla bla
```

The hurdles I see are:

- Make runs each command in a subshell, which is great for a build system, but not great for a collection of scripts (like what we're using it for), so it might be hard to export these variables like you've seen earlier.
- If we tried to avoid environment variables by passing the files to the appropriate tools (which is good IMO, envvars are a bit too fiddly), I don't know what format would work for Terraform and the shell. Terraform seems to have godawful support for .env files.

Let's think about this a bit more, but if we can't come up with anything better, I think this approach is still workable.

.env.staging
```diff
@@ -0,0 +1,2 @@
+export ENVIRONMENT=staging
```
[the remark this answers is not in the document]
chaitanyabaranwal: For some reason removing `export` the environment variables aren't being properly loaded. We can talk about this more but keeping the `export` for now seems fine to me.

chaitanyabaranwal:
@glencbz I think solution 1 (of including `source ../../.env.staging` will not work, because Makefiles load the entire command then execute it (if I understand things correctly). So if we do something like

```
init:
    source ../../.env.staging && cd $(ENVIRONMENT) and terraform init
```

It will actually translate to:

```
source ../../.env.staging && cd <empty> and terraform init
```

This is because the environment variables are not loaded _before_ executing the `make` command.

Regarding the second solution, I think there is a way to make things work for Terraform ([Stackoverflow link](https://stackoverflow.com/questions/59789730/how-can-i-read-environment-variables-from-a-env-file-into-a-terraform-script)). But I am worried about the frontend deployment because the frontend `Makefile` still needs the bucket link somehow. What do you think?

glencbz:
The approach is under active discussion on Slack, so lots of things could change. In the meantime, I'll respond to some specific comments..

> I think solution 1 (of including source ../../.env.staging will not work, because Makefiles load the entire command then execute it

In the example, `ENVIRONMENT` is a Make variable, not an environment variable, so I think this should have worked.

> Regarding the second solution, I think there is a way to make things work for Terraform ([Stackoverflow link](https://stackoverflow.com/questions/59789730/how-can-i-read-environment-variables-from-a-env-file-into-a-terraform-script)). But I am worried about the frontend deployment because the frontend Makefile still needs the bucket link somehow. What do you think?

That could work, the code is not too hard to understand and it does solve the problem for Terraform. I worry a bit whether we can extend this in the future (e.g. if we need to relocate the `.env` file), since scripting in Terraform is quite limited (and rightfully so, since its primary purpose is readability).

On Slack, @chaychoong suggested some approaches to getting envvars in the Makefile using `sh -c 'source env-file; real-command'`.

chaitanyabaranwal:
@glencbz Made the latest changes in https://github.com/bettersg/call-home/pull/119/commits/38731e489fdd939058d59be6831814832945aea3. Sorry for the delay!

infra/Makefile
```diff
@@ -1,7 +1,7 @@
 ENVIRONMENT?=staging
 
 gcloud-login:
-	cd $(ENVIRONMENT) && gcloud auth application-default login
```
glencbz, line 4: What's the reasoning behind this change? Cleanup of unnecessary code?

infra/staging/main.tf
```diff
@@ -32,6 +32,11 @@ variable "db_password" {
   sensitive = true
 }
 
+variable "frontend_bucket_name" {
```
glencbz, line 35: Hm, relying on default values isn't great for reproducibility, so at some point I think we'd want something like `variables.tfvars`. Your change is in line with the existing code though, so let's not change it. We'll just keep this in mind as a TODO (or even better, we could leave a TODO comment).

infra/staging/main.tf
```diff
@@ -32,6 +32,11 @@ variable "db_password" {
...
+  type = string
+  default = "call-home-frontend"
```
glencbz, line 37: There is a good chance that we will have to rename this to something like `call-home-frontend-staging`, since we'll have a prod instance too, but we can cross that bridge when we get there.

infra/README.md
```diff
@@ -6,10 +6,14 @@ Ensure that you have `terraform` and `gcloud` installed locally.
 
 Run `make gcloud-login` to login using `gcloud` if you are not logged in already.
 
+### Using Terraform
+
+Make sure you have the appropriate environment variables set using `source .env.[environment]` from the root folder.
```
glencbz, line 11: Is this still accurate? IIRC we wanted any `source`-ing (if any) to happen in the `make` recipe using the `sh -c` trick.

apps/super-app/README.md
```diff
@@ -39,6 +39,12 @@ Instead, it will copy all the configuration files and the transitive dependencie
 
 You don’t have to ever use `eject`. The curated feature set is suitable for small and middle deployments, and you shouldn’t feel obligated to use this feature. However we understand that this tool wouldn’t be useful if you couldn’t customize it when you are ready for it.
 
+## Deployment
+
+* Use `make build` to create the production build.
+* After creating the build, you can load the appropriate environment variables set `source .env.[environment]` from the root folder.
```
glencbz, line 45: Is this still accurate? It seems like this is obsolete with the `sh -c` Makefile trick

apps/super-app/README.md
```diff
@@ -39,6 +39,12 @@ Instead, it will copy all the configuration files and the transitive dependencie
...
+* Run `make frontend-deploy` to deploy the production build to a GCS bucket.
```
glencbz, line 46: s/make frontend-deploy/make deploy

glencbz (approved):
It looks like some bits of documentation are out of date, do take a look at them before merging.
Otherwise, the core changes look good, so feel free to make the cleanups in your own time and merge them :) I only ask that you use the "squash and merge" button so that we don't see the commits that are just fixing up PR issues.

PR #121, opened 2023-02-12
Add service card component

stevahnes:
Fixes https://github.com/bettersg/call-home/issues/125

Plan is to use MUI Card to create the different services' reusable cards. These cards should be able to take it the values from a separate service. Then modify Scratch to Home Page and add the card as landing page.

image


apps/super-app/src/services/SupportServices.ts
```diff
@@ -33,6 +41,21 @@ export interface FacebookLink {
   text: string;
 }
 
+export function getServiceCardAbouts(): ServiceCardAbout[] {
```
glencbz, line 49: I think this should go into another file. This is becoming a catch-all for all info coming from Strapi, but this file was intended to be specific to the support detail page. There's going to be some work required to split up the support detail bits from the TWC2 bits, and adding more things will make that process even more complicated.

glencbz:
Looks promising so far :)

apps/super-app/src/services/SupportServices.ts
stevahnes: Gotcha. I think we should reuse the root interface and extend it to both SupportDetail and ServiceCard interfaces, what do you think?

glencbz: Discussed on Slack, we agreed to move the 'common' things to a separate object (not a root interface).

apps/super-app/src/services/SupportServices.ts
```diff
@@ -1,10 +1,23 @@
 // TODO This is a stub for testing, eventually replace this with Strapi Calls
+
+import { Path } from '../routes/paths';
```
glencbz: Ah, I hadn't considered the problem of adding links to this page.. I think this is okay for now, but we have no good way for the actual Strapi service to know what the app internal constants are, so when we move to Strapi, we will have to drop this import. Let's leave a TODO to note that we need to figure this out later.

I'm quite certain that the backend shouldn't be able to tell the frontend what the route is, and we need to create each support detail page anyway, so and it doesn't make sense for the backend to tell the frontend what the list of support services is.

I'm not sure what the 'right' way forward is, here are some ideas:

- Keep the `getServiceCardAbouts()`, and teach the frontend to identify the route that should correspond to a service. (e.g. something like associate each support service entry with an 'id' and store a map of support service id to route in the frontend). Seems a little roundabout, but this maps well onto Strapi content types. I worry about the id being accidentally modified in Strapi and breaking the frontend, but maybe there are ways to make this less likely.
- Get rid of `getServiceCardAbouts()` and hardcode the list of support services in the frontend. Since the frontend should already know the list of services, it might make sense to just have the frontend reach out to the right backend routes. But if we do this, it probably makes the most sense for to have only one-off types for each Support service (e.g. `/pbs` and `/twc2` instead of `/support/1` and `/support/2`), and I don't know how this would map well onto Strapi.

WDYT? At any rate, we don't have to make a decision right now, it seems like this would change a lot based on how we experiment with Strapi.

apps/super-app/src/common/components/ServiceCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export function ServiceCard(props: ServiceCardAbout) {
+  const { logo, name, shortBlurb, route } = props;
+  return (
+    <Card sx={{ display: 'flex', marginBottom: '16px' }}>
+      <CardMedia
+        component="img"
+        sx={{ width: '5rem', height: '5rem', margin: 'auto 16px' }}
+        image={logo}
+        alt="TWC2 logo"
```
glencbz, line 21: Shouldn't this come from the props too? At any rate, I'm also okay to omit the `alt`, I don't think it will matter in practice.

glencbz (changes requested):

apps/super-app/src/common/components/ServiceCard.tsx
stevahnes: HAHA GOOD CATCH

apps/super-app/package.json
```diff
@@ -27,8 +27,8 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "lint": "eslint .",
-    "format": "prettier --write 'src/**/*.{tsx,js,ts}'",
-    "format:check": "prettier --check 'src/**/*.{tsx,js,ts}'"
+    "format": "prettier --write src/**/*.{tsx,js,ts}",
+    "format:check": "prettier --check src/**/*.{tsx,js,ts}"
```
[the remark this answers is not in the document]
stevahnes: @glencbz the command didn't work for me when there's `'`. Were they added for a specific purpose?

apps/super-app/src/common/components/ServiceCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+          <CardContent>
+            <Typography
+              component="div"
+              variant="h5"
+              sx={{
+                fontSize: 16,
+                fontWeight: 600,
+                color: '#297B94',
+                fontFamily: 'Hind',
+                textDecoration: 'underline',
```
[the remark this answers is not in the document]
stevahnes: @glencbz just haven't figured out the structure of the classes yet :(

apps/super-app/src/routes/paths.ts
```diff
@@ -1,6 +1,7 @@
 export enum Path {
   Home = '/',
   // TODO these are just for testing. Delete this before launch.
-  Scratch = '/scratch',
+  LandingPage = '/landing-page',
   SupportDetail = '/support-detail',
+  ProBonoSG = '',
```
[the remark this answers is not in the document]
stevahnes: Placeholder, okay let's make it `pro-bono`.

apps/super-app/src/routes/LandingPage.tsx
```diff
@@ -1,24 +1,36 @@
...
     <Container
       style={{
         background: 'no-repeat url(/images/background.svg) bottom center',
         backgroundSize: 'contain',
+        padding: '0px 20px',
       }}
     >
+      <Typography
+        component="div"
+        variant="h5"
```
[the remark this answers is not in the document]
stevahnes: Turned into TODO

apps/super-app/src/common/components/ServiceCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+        image={logo}
+        alt="TWC2 logo"
+      />
+      <CardActionArea component={Link} to={route}>
+        <Box sx={{ flex: '1 0 auto' }}>
+          <CardContent>
+            <Typography
+              component="div"
+              variant="h5"
+              sx={{
```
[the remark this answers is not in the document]
stevahnes: @glencbz yea, ehm actually I need to learn more about this. Felt a bit lost with MUI, so apologies for the mess in the styling!

apps/super-app/src/services/SupportServices.ts
stevahnes: Let's do it as a TODO first?

apps/super-app/src/routes/LandingPage.tsx
```diff
@@ -0,0 +1,44 @@
...
+        sx={{
+          fontSize: 20,
+          fontWeight: 700,
+          color: '#133844',
+          fontFamily: 'Hind',
+          lineHeight: '28px',
+          margin: '16px 0px',
+        }}
+      >
+        Free Support Services
```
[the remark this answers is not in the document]
stevahnes: Resolved. We have a TODO for returning something still as the page now won't render at all if we're getting this string from Strapi and it's down.

apps/super-app/src/common/components/ServiceCard.tsx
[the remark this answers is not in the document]
stevahnes: I moved it to theme for now!

apps/super-app/src/services/ViewServices.ts
```diff
@@ -45,6 +54,25 @@ export interface FacebookLink {
   text: string;
 }
 
+export function getServiceCardAbouts(): ServiceCardAbout[] {
```
stevahnes, line 57: @glencbz technically this is not a "common" function as it's specific to the Service Cards.

apps/super-app/src/common/components/ServiceCard.tsx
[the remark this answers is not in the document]
stevahnes: I moved it to theme for now!

apps/super-app/package.json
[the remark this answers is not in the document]
glencbz: Adding `'` quotes would prevent the shell from expanding the glob pattern instead of passing the glob directly to `prettier`. I don't recall if this is actually needed or not, but it seemed more portable to me to rely on the tool's globbing instead of the shell's.

What platform are you running this on, and what is the error message you get? Also, did you invoke this via `make format`?

glencbz:
Btw, if you are comfortable force-pushing to GitHub, are you comfortable with rewriting the commits to create a clean history instead of pushing PR fixups as new commits? It's okay if not, it's mostly a matter of personal preference - I find it easier to review a clean history since I can review each commit individually instead of a whole PR.

apps/super-app/package.json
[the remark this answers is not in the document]
stevanussatria: I was using `non run format` and it didn't run 😅 I'll remove it first if it's gonna affect CI, and try with `make`!

stevanussatria:
> Btw, if you are comfortable force-pushing to GitHub, are you comfortable with rewriting the commits to create a clean history instead of pushing PR fixups as new commits? It's okay if not, it's mostly a matter of personal preference - I find it easier to review a clean history since I can review each commit individually instead of a whole PR, and we can have each clean commit in the merged history.

Should I squash it all to one commit then?

glencbz:
> Should I squash it all to one commit then?

That's also okay. We could also just use the "squash" button on GitHub.

apps/super-app/src/App.test.tsx
```diff
@@ -3,27 +3,30 @@ import { render, screen } from '@testing-library/react';
 import App from './App';
 import { MemoryRouter } from 'react-router-dom';
 
-test('renders a nonsense homepage', () => {
+test('renders LandingPage component on load homepage', () => {
```
glencbz, line 6: > `renders LandingPage component on load homepage`

This doesn't read right. For consistency with the other test maybe `renders LandingPage component on /`?
Also, the test isn't actually asserting that the component is being rendered, but that the expected elements are on the page, so maybe something like `renders support cards on /`?

apps/super-app/src/App.test.tsx
```diff
@@ -3,27 +3,30 @@ import { render, screen } from '@testing-library/react';
...
+  const header = screen.getByText('Free Support Services');
+  expect(header).toBeInTheDocument();
+  const twc2 = screen.getByText('Transient Workers Count Too');
+  expect(twc2).toBeInTheDocument();
+  const pbsg = screen.getByText('Pro Bono SG');
+  expect(pbsg).toBeInTheDocument();
 });
 
-test('renders some nonsense buttons on /scratch', () => {
+test('make sure app initializes on /landing-page', () => {
```
glencbz, line 20: The pre-established style is to use the singular present tense as if the codebase is doing something, e.g. `(it) renders some buttons`, `(it) should do X`. I think it might have been part of [mocha](https://mochajs.org/#assertions)'s legacy, but I've seen the `should do X when Y` pattern in quite a few places tbh.

apps/super-app/src/common/components/ServiceCard.tsx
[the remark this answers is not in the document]
glencbz: I see that there were changes in `Theme.ts`, but there are still some overriding styles here. Is this intended?

apps/super-app/src/common/components/ServiceCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+            </Typography>
+            <Typography
+              variant="subtitle1"
+              color="text.secondary"
+              component="div"
+              sx={{
+                fontSize: 14,
+                fontWeight: 400,
+                color: '#000000',
+                fontFamily: 'Hind',
```
glencbz, line 47: Similarly, I'm not sure if you meant for this to be on the component or in the theme.

apps/super-app/src/routes/LandingPage.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export function LandingPage() {
+  const [fixedStrings, setFixedStrings] = useState<SupportDetailStrings | null>(
+    null
+  );
+  useEffect(() => {
+    setFixedStrings(getSupportDetailStrings());
+  }, []);
+  if (!fixedStrings) {
+    // TODO return something (hint) since this is landing page, we can't return null
```
glencbz, line 21: We'll probably render a loading spinner :)

apps/super-app/src/services/SupportAboutServices.ts
```diff
@@ -0,0 +1,6 @@
+export interface SupportAbout {
```
glencbz, line 1: > apps/super-app/src/services/SupportAboutServices.ts

I don't quite understand the file name. Is this a (software) service that serves `SupportAbout`s? If so, then let's drop the `Services` suffix; it's redundant with the directory name. In either case, do stick with the singular noun for the file name.

Maybe this is a (real life service), but I'm not sure how to interpret the name.

apps/super-app/src/services/ViewServices.ts
```diff
@@ -1,5 +1,18 @@
 // TODO This is a stub for testing, eventually replace this with Strapi Calls
```
glencbz, line 1: > apps/super-app/src/services/ViewServices.ts

Similarly, I'm not sure how to read this file name. Is this a service that serves up `View`s? (I don't know what a View is).

apps/super-app/package.json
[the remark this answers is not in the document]
glencbz: > What platform are you running this on, and what is the error message you get?

What are the answers to these questions? I can help troubleshoot.

>  I'll remove it first if it's gonna affect CI, and try with make

I don't think it'll affect CI, but yes, let's remove it since we don't really understand why we need to make this change right now.

glencbz (changes requested):
Thanks for your patience, I only had small nitpicky comments this time, so this is pretty close to merging.

apps/super-app/package.json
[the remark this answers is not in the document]
stevahnes: > What are the answers to these questions? I can help troubleshoot.

`npm run format` ah. Directly in bash.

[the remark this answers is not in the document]
glencbz: > What platform are you running this on, and what is the error message you get?

I'd still like to know the answers to these questions. `npm run format ah. Directly in bash.` doesn't give me enough troubleshooting info.

> I don't think it'll affect CI, but yes, let's remove it since we don't really understand why we need to make this change right now.

We wanted to undo this change, right?

apps/super-app/src/App.test.tsx
```diff
@@ -3,27 +3,30 @@ import { render, screen } from '@testing-library/react';
...
+  const header = screen.getByText('Free Support Services');
+  expect(header).toBeInTheDocument();
+  const twc2 = screen.getByText('Transient Workers Count Too');
+  expect(twc2).toBeInTheDocument();
+  const pbsg = screen.getByText('Pro Bono SG');
+  expect(pbsg).toBeInTheDocument();
 });
 
-test('renders some nonsense buttons on /scratch', () => {
+test('it should render support cards on /landing-page', () => {
```
glencbz, line 20: Ah, I didn't mean that we should prepend the names with `it`, I meant that we should write the name _as if_ there was an "it" in front. So in these cases, the fix would be to remove the "it".

apps/super-app/src/common/Theme.ts
```diff
@@ -30,6 +30,26 @@ const callHomeTheme = createTheme(
...
+        color: '#133844',
+        fontFamily: 'Hind',
+        lineHeight: '28px',
+      },
+      h6: {
+        fontSize: 16,
+        fontWeight: 600,
+        color: '#297B94',
+        fontFamily: 'Hind',
+        textDecoration: 'underline',
```
glencbz, line 45: I wonder if we'll want all `h6`s to be underlined. Let's keep it and see if this ages well.

apps/super-app/src/routes/LandingPage.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export function LandingPage() {
+  const [fixedStrings, setFixedStrings] = useState<SupportDetailStrings | null>(
+    null
+  );
+  useEffect(() => {
+    setFixedStrings(getSupportDetailStrings());
+  }, []);
+  if (!fixedStrings) {
+    // TODO return something (hint or loading spinner) since this is landing page, we can't return null
```
glencbz, line 21: IMO returning `null` is ok, since it'll just look like the page is loading. But a spinner is nicer, yes :)

glencbz (approved):
I think this is a step in the right direction, so I'm ok to merge it right now.

For cleanliness, I would prefer to make the changes requested `apps/super-app/package.json` and `apps/super-app/src/App.test.tsx`, but it's close enough that I won't insist on it.

Since this has approval, you can choose to make the changes (or not) and push the merge button. If you choose to merge this, please use the 'squash and merge' option so that we don't see the messy intermediate merges and fixups.

apps/super-app/src/App.test.tsx
stevahnes: Oh right now I rmb, the Angular Mocha/Jasmine syntax was `it(_______)`

apps/super-app/src/routes/LandingPage.tsx
stevahnes: Let's keep it as TODO

stevahnes:
@glencbz wanted to fix it but my other GitHub account ain't no access lol. I'll fix the comments in a diff PR!

PR #135, opened 2023-02-27
personal-calls: Verify FIN

glencbz:
Fixes https://github.com/bettersg/call-home/issues/140

This will be our replacement for verifying work pass numbers, at least in the interim.

This PR does the following:
* Add FIN verification, largely by copying the existing logic for verifying work pass numbers on the frontend and backend
* Lightly refactor the checks for whether a user is verified so that it's easier to add and remove verification conditions.
* Add feature flags to disable workpass verification and to enable fin verification

### DIFF
--- a/personal-calls/frontend/src/services/Feature.ts
+++ b/personal-calls/frontend/src/services/Feature.ts
@@ -9,6 +9,8 @@ export interface FeatureState {
   EDGE_EXPERIMENT: Edge;
   FEEDBACK_DIALOG: boolean;
   CREDIT_CAP: boolean;
+  DISABLE_WORKPASS: boolean;
+  ENABLE_FIN: boolean;
 }
 
 const featureEndpoint = '/features';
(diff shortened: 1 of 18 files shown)


personal-calls/frontend/src/services/Feature.ts
```diff
@@ -21,6 +23,8 @@ export default class FeatureService extends ObservableService<FeatureState> {
       EDGE_EXPERIMENT: Edge.DEFAULT,
       FEEDBACK_DIALOG: false,
       CREDIT_CAP: false,
+      DISABLE_WORKPASS: false,
+      ENABLE_FIN: false,
```
stevanussatria, line 27: We're not enabling this here is it?

glencbz: These are just default values. We enable flags with environment variables. All the flags should default to false, meaning "flag was not given in the environment"

glencbz:
@stevahnes if this looks good to you, could you approve this PR?

chaitanyabaranwal (approved):
It looks good to me but my review must be taken with a pinch of salt since I don't have much Typescript/React experience :) 

@glencbz Trust that you've tested this out properly!

glencbz:
Thanks! Feel free to ask questions about the patterns and how things are implemented, e.g. I'm considering using a similar approach for feature flags for the Call Home super app.

PR #137, opened 2023-03-02
TWC2 add whatsapp deep link

e056:
Fixes https://github.com/bettersg/call-home/issues/131

https://github.com/bettersg/call-home/issues/131#issuecomment-1445155469
image

PR #144, opened 2023-03-10
TWC2 add whatsapp deep link

### LINKED ISSUE #131: [TWC2] Add Whatsapp deep link
For the TWC2 sub app, we want that CTA button to open TWC2's WhatsApp. Figure out how to do this.

Create a React component that renders a `button` that links to WhatsApp when clicked. The WhatsApp url should be the http://wa.me one that opens the WhatsApp app if it is installed, and prompts to install if not.

e056:
Fixes #131 

- deleted previous [PR](https://github.com/bettersg/call-home/pull/137) 
- componentize button (to do)
image


apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -115,16 +115,23 @@ function DescriptionSection(props: DescriptionProps) {
...
   return (
     <Box className="support-cta-container" sx={{ borderTopColor: 'grey.200' }}>
       <div style={{ flex: 4, height: '100%' }}>{ctaBlurb}</div>
       <PrimaryButton
+        href={ctaLink}
+        target="_blank"
+        rel="noopener"
         sx={{ flex: 5, backgroundColor: 'primary.700', height: '100%' }}
       >
+        <WhatsappIcon></WhatsappIcon>
```
stevahnes, line 134: @e056 I am not sure if it's possible but can we explore switching up the icon as well based on some props?

For example, for Pro Bono SG we will use a schedule (calendar) icon instead. If not then that's fine too!

I think what we can do is have a prop for `icon`, and then do a conditional within the button to choose which button to render. @glencbz any thoughts?

apps/super-app/src/services/DetailServices.ts
```diff
@@ -95,5 +96,6 @@ export function getTwc2Detail(): Twc2SupportDetail {
     },
     ctaBlurb: 'Call on Mondays to Fridays (9am - 9pm), message any time',
     ctaButtonText: '+65 6297 7564',
+    ctaLink: 'https://wa.me/6583668579',
```
stevahnes, line 99:
```suggestion
    ctaLink: 'https://wa.me/6562977564',
```

stevahnes (changes requested):

apps/super-app/src/routes/SupportDetailWip.tsx
e056: Ahh okie sure, I will look into that

glencbz: I think it's okay not to worry about it for now, we can alway refactor out the commonality if we need to (we may not necessarily need it).

stevahnes: Cool! Later is fine too.

apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -115,16 +115,23 @@ function DescriptionSection(props: DescriptionProps) {
...
-  const { ctaBlurb, ctaButtonText } = props;
+  const { ctaBlurb, ctaButtonText, ctaLink } = props;
   return (
     <Box className="support-cta-container" sx={{ borderTopColor: 'grey.200' }}>
       <div style={{ flex: 4, height: '100%' }}>{ctaBlurb}</div>
       <PrimaryButton
+        href={ctaLink}
+        target="_blank"
+        rel="noopener"
         sx={{ flex: 5, backgroundColor: 'primary.700', height: '100%' }}
```
stevahnes, line 132: @e056 should we move this into theme? CC: @glencbz

stevahnes (approved):

apps/super-app/src/routes/SupportDetailWip.tsx
e056: Hi @stevahnes , sorry for the late reply. Sorry which part are you referring to?

stevahnes: The `sx` portion

apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -14,7 +14,7 @@ import {
 import { Container } from '../common/components';
 import { PrimaryButton } from '../common/components/RoundedButton';
 import './SupportDetailWip.css';
-
+import { ReactComponent as WhatsappIcon } from './../common/whatsapp-icon.svg';
```
glencbz, line 17: Unless there's a valid reason, do try to avoid changing spacing.

apps/super-app/src/routes/SupportDetailWip.tsx
glencbz: I don't think you meant theme? The `sx` is already referencing theme variables (e.g. `primary.700`).

Perhaps you meant making this a shareable component. That will be good to do someday, but that's a refactor that's out of scope for this feature.

glencbz (changes requested):
`apps/super-app/src/common/whatsapp-icon.svg`

The common directory isn't the right place for this. From `apps/super-app/src/common/README.md`:

> This contains files that _should_ be shared between apps, but we just haven't
for whatever reasons, e.g.:

This is specific to the `super-app/`, so it should go into another directory. I suggest either adding it to `public/images` (and use an `` with the right public url, see how other things in `public/images` are used), or creating a new component for the CTA button and putting the image in the component's directory (like you did with the navbar).

apps/super-app/src/routes/SupportDetailWip.tsx
stevahnes: Hmmm, I initially thought that we should probably define a theme for all primary buttons instead? Unless we foresee a different primary button look and feel?

apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -115,16 +115,23 @@ function DescriptionSection(props: DescriptionProps) {
...
+>;
 
 function CtaSection(props: CtaProps) {
-  const { ctaBlurb, ctaButtonText } = props;
+  const { ctaBlurb, ctaButtonText, ctaLink } = props;
   return (
     <Box className="support-cta-container" sx={{ borderTopColor: 'grey.200' }}>
       <div style={{ flex: 4, height: '100%' }}>{ctaBlurb}</div>
       <PrimaryButton
+        href={ctaLink}
```
glencbz, line 129: I was very surprised that this works, since this isn't supported on native HTML buttons. It turns out that Material UI buttons actually render a `<a>` if you pass the `href` prop (https://mui.com/material-ui/api/button/#props). Wow.

apps/super-app/src/routes/SupportDetailWip.tsx
glencbz: Hm, I don't follow. The word 'theme' sounds like you're referring to Material UI themes (https://mui.com/material-ui/customization/theming/), but that can't be what you mean.

It sounds like you mean custom styles and a custom component (which is what I described). In which case, yes we do want a fixed set of styles for our CTA button. But @e056 isn't adding a new instance of the button or making any substantial change that warrants us refactoring the code yet, so I don't think that needs to be done noow.

stevahnes (approved):
LGTM!

stevahnes (approved):

apps/super-app/src/services/DetailServices.ts
```diff
@@ -97,5 +98,6 @@ export function getTwc2Detail(): Twc2SupportDetail {
     ctaBlurb: 'Call on Mondays to Fridays (9am - 9pm), message any time',
     ctaButtonText: '+65 6297 7564',
     ctaLink: 'https://wa.me/6562977564',
+    whatsappLogoSrc: '/images/whatsapp-icon.svg',
```
glencbz, line 101: Hm, I think we should just reference the `src` string from the component since it's not like this property is going to come from Strapi.

We can clean this up later, but do leave a TODO comment if so.

glencbz (approved):
LGTM, but let's do something about about the whatsapp logo src before merging this.  As long as we make the change, I think it's okay for anyone to merge this, e.g. @stevahnes you can do it if it will make your life easier working on your other PR.

Since we're going to squash and merge, I think it's fine to add another commit on top to add a TODO/hardcode the src prop.

apps/super-app/src/services/DetailServices.ts
stevahnes: @glencbz what do you mean by this exactly? Understand that the image will be part of the FE asset instead, but what do you mean by `src` string from component?

apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -115,16 +115,24 @@ function DescriptionSection(props: DescriptionProps) {
...
+        target="_blank"
+        rel="noopener"
         sx={{ flex: 5, backgroundColor: 'primary.700', height: '100%' }}
       >
+        <img src={whatsappLogoSrc} />
```
glencbz, line 134: @stevahnes I meant that it's okay to hardcode the value here. Also, I would appreciate it if you wait for my clarification before merging the PR when I set specific guidelines around merging it.

stevahnes: Yeah, well you said I can merge it and work on it on my PR, so that's what I am doing @glencbz. Also, it was my suggestion to not hardcode the value because if you see the other mocks, we are using other icons for other CTA buttons.

glencbz: Sorry, I didn't mean to come across as overriding your suggestion.

What I mean is that if we have a component called `WhatsappIcon`, it doesn't make sense to parameterise the URL for the WhatsApp image (especially if it's also called whatsappLogoSrc) because we're only ever going to have one anyway.

If we want a generic CTA that can accept any image URL, then sure, parameterising is fine. Either approach can work as long as it's coherent.

PR #145, opened 2023-03-12
Add HTTP(S) load balancer for GCS bucket

chaitanyabaranwal:
Fixes https://github.com/bettersg/call-home/issues/150

This PR adds additional resources to the staging Terraform script, with the main goal of enabling serving the frontend (from a GCS bucket) over a static IP address. I used the following guide: https://cloud.google.com/cdn/docs/setting-up-cdn-with-bucket. 

Details on the resources added:
* Compute Engine API service (needed for creating IP address)
* Static IP address for load balancer
* Backend bucket to run the load balancer in
* URL map for the load balancer
* SSL certificate creation for HTTPS capabilities
* HTTPS proxy for the load balancer
* Forwarding rule for the load balancer

### DIFF
--- a/infra/staging/main.tf
+++ b/infra/staging/main.tf
@@ -39,6 +39,11 @@ variable "frontend_bucket_name" {
   default = "call-home-frontend"
 }
 
+variable "website_domain_name" {
+  type = string
+  default = "app2-staging.callhome.sg"
+}
+
 ######################
 ## Project Settings ##
 ######################
@@ -201,7 +206,7 @@ resource "google_compute_managed_ssl_certificate" "call_home_frontend_lb_ssl" {
   project = data.google_project.project.project_id
 
   managed {
-    domains = ["app2-staging.callhome.sg"]
+    domains = [var.website_domain_name]
   }
 }
 
--- a/infra/staging/main.tf
+++ b/infra/staging/main.tf
@@ -189,8 +189,8 @@ resource "google_compute_backend_bucket" "call_home_frontend_lb" {
 }
 
 // URL map for the backend bucket
-resource "google_compute_url_map" "call_home_frontend_lb_url_map" {
-  name = "call-home-frontend-lb-url-map"
+resource "google_compute_url_map" "call_home_lb" {
+  name = "call-home-lb"
   project = data.google_project.project.project_id
   default_service = google_compute_backend_bucket.call_home_frontend_lb.id
 }
@@ -213,7 +213,7 @@ resource "google_compute_managed_ssl_certificate" "call_home_frontend_lb_ssl" {
 resource "google_compute_target_https_proxy" "call_home_frontend_lb_proxy" {
   name = "call-home-frontend-lb-proxy"
   project = data.google_project.project.project_id
-  url_map = google_compute_url_map.call_home_frontend_lb_url_map.id
+  url_map = google_compute_url_map.call_home_lb.id
 
   ssl_certificates = [
     google_compute_managed_ssl_certificate.call_home_frontend_lb_ssl.name
--- a/infra/staging/main.tf
+++ b/infra/staging/main.tf
@@ -193,6 +193,10 @@ resource "google_compute_url_map" "call_home_lb" {
   name = "call-home-lb"
   project = data.google_project.project.project_id
   default_service = google_compute_backend_bucket.call_home_frontend_lb.id
+
+  depends_on = [
+    google_compute_backend_bucket.call_home_frontend_lb
+  ]
 }
 
 // Setup SSL certificate to enable HTTPS for load balancer
@@ -220,7 +224,8 @@ resource "google_compute_target_https_proxy" "call_home_frontend_lb_proxy" {
   ]
 
   depends_on = [
-    google_compute_managed_ssl_certificate.call_home_frontend_lb_ssl
+    google_compute_managed_ssl_certificate.call_home_frontend_lb_ssl,
+    google_compute_url_map.call_home_lb
   ]
 }
 
@@ -233,6 +238,11 @@ resource "google_compute_global_forwarding_rule" "call_home_frontend_lb_forwardi
   port_range = "443"
   target = google_compute_target_https_proxy.call_home_frontend_lb_proxy.id
   ip_address = google_compute_global_address.call_home_frontend_ip.id
+
+  depends_on = [
+    google_compute_target_https_proxy.call_home_frontend_lb_proxy,
+    google_compute_global_address.call_home_frontend_ip
+  ]
 }
 
 #############
--- a/infra/staging/main.tf
+++ b/infra/staging/main.tf
@@ -151,6 +151,61 @@ resource "google_storage_bucket_iam_member" "frontend_bucket_viewer" {
   role   = "roles/storage.objectViewer"
   member = "allUsers"
 }
+#######################################
+## Load balancer for frontend bucket ##
+#######################################
+
+// NOTE: Had to enable Compute Engine API on the project for this to work
+// Following the guide here: https://cloud.google.com/cdn/docs/setting-up-cdn-with-bucket
+
+// Setup IP address for load balancer
+resource "google_compute_global_address" "call_home_frontend_ip" {
+  project = data.google_project.project.project_id
+  name = "call-home-frontend-ip"
+}
+
+// Backend bucket with CDN policy with default ttl settings
+resource "google_compute_backend_bucket" "call_home_frontend_lb" {
+  name = "call-home-frontend-lb"
+  description = "Load balancer for the frontend GCS bucket"
+  project = data.google_project.project.project_id
+  bucket_name = google_storage_bucket.call_home_frontend.name
+  enable_cdn = true
+
+  cdn_policy {
+    cache_mode = "CACHE_ALL_STATIC"
+    client_ttl = 3600
+    default_ttl = 3600
+    max_ttl = 86400
+    negative_caching = true
+    serve_while_stale = 86400
+  }
+}
+
+// URL map for the backend bucket
+resource "google_compute_url_map" "call_home_frontend_lb_url_map" {
+  name = "call-home-frontend-lb-url-map"
+  project = data.google_project.project.project_id
+  default_service = google_compute_backend_bucket.call_home_frontend_lb.id
+}
+
+// HTTP proxy
+resource "google_compute_target_http_proxy" "call_home_frontend_lb_proxy" {
+  name = "call-home-frontend-lb-proxy"
+  project = data.google_project.project.project_id
+  url_map = google_compute_url_map.call_home_frontend_lb_url_map.id
+}
+
+// Forwarding rule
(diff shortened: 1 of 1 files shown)


infra/staging/main.tf
```diff
@@ -151,6 +151,66 @@ resource "google_storage_bucket_iam_member" "frontend_bucket_viewer" {
...
+    serve_while_stale = 86400
+  }
+}
+
+// URL map for the backend bucket
+resource "google_compute_url_map" "call_home_frontend_lb_url_map" {
+  name = "call-home-frontend-lb-url-map"
+  project = data.google_project.project.project_id
+  default_service = google_compute_backend_bucket.call_home_frontend_lb.id
+}
```
glencbz, line 195: @halfwhole @chaitanyabaranwal I suspect the url map is how we'll end up connecting Strapi to the load balancer. It seems likely that we'll add some rules to the url map (https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_url_map#example-usage---url-map-bucket-and-service) to direct different urls to different backends - either the GCS bucket or the App Engine app.

e.g.I think we might want something like

- Direct `app2[-staging].callhome.sg` to the GCS bucket
- Direct `app2[-staging]-api.callhome.sg` or `app2[-staging].callhome.sg/api` to the Strapi App Engine app.

which _maybe_ translates into something like 

```
  host_rule {
    hosts        = ["app2-staging.callhome.sg"]
    path_matcher = "callhome"
  }
  path_matcher {
    name            = "callhome"
    default_service = my_bucket.static.id

    path_rule {
      paths   = ["/api"]
      service = my_app_engine.id
    }
  }
```

infra/staging/main.tf
```diff
@@ -151,6 +151,66 @@ resource "google_storage_bucket_iam_member" "frontend_bucket_viewer" {
...
+    default_ttl = 3600
+    max_ttl = 86400
+    negative_caching = true
+    serve_while_stale = 86400
+  }
+}
+
+// URL map for the backend bucket
+resource "google_compute_url_map" "call_home_frontend_lb_url_map" {
+  name = "call-home-frontend-lb-url-map"
```
glencbz, line 192: Confusingly, this maps to two places  in the GCP console:

- Load balancer: https://console.cloud.google.com/net-services/loadbalancing/details/http/call-home-frontend-lb-url-map?project=call-home-staging
- URL map field in https://console.cloud.google.com/net-services/loadbalancing/advanced/targetHttpProxies/details/call-home-frontend-lb-proxy?project=call-home-staging

Keeping the `url-map` name is okay, though it is a bit weird that the Load Balancer is named `*-url-map`. Maybe we should rename this to `call-home-frontend-lb` or `call-home-lb` (if we think we will reuse it for the Strapi App Engine app).

If we rename it, take note that this sort of networking resource is a bit finicky to delete in Terraform because networking requires a lot of external interactions and resource creation/teardown take a long time. Fortunately, we can change the internal Terraform address using `terraform state mv` (https://stackoverflow.com/questions/70418172/terraform-rename-a-resource-without-deleting) and I _think_ we can rename the `.name` property. (For this reason, it doesn't matter if we name it either `call-home-frontend-lb` or `call-home-lb`, since we can rename it afterwards.)

infra/staging/main.tf
```diff
@@ -151,6 +151,66 @@ resource "google_storage_bucket_iam_member" "frontend_bucket_viewer" {
...
+
+// Setup IP address for load balancer
+resource "google_compute_global_address" "call_home_frontend_ip" {
+  project = data.google_project.project.project_id
+  name = "call-home-frontend-ip"
+}
+
+// Backend bucket with CDN policy with default ttl settings
+resource "google_compute_backend_bucket" "call_home_frontend_lb" {
+  name = "call-home-frontend-lb"
```
glencbz, line 174: It would be more accurate to call this the `backend_bucket`, since that's what Terraform and the GCP console UI call it https://console.cloud.google.com/net-services/loadbalancing/advanced/backendBuckets/details/call-home-frontend-lb?project=call-home-staging

infra/staging/main.tf
```diff
@@ -151,6 +151,66 @@ resource "google_storage_bucket_iam_member" "frontend_bucket_viewer" {
...
   member = "allUsers"
 }
+#######################################
+## Load balancer for frontend bucket ##
+#######################################
+
+// Following the guide here: https://cloud.google.com/cdn/docs/setting-up-cdn-with-bucket
+
+// Enable Compute Engine API to setup an IP address
+resource "google_project_service" "compute_engine_api" {
```
glencbz, line 161: I don't know why exactly this is needed, but I'm willing to accept that it's needed to setup an IP address _just because_. GCP has a bunch of weird legacy problems with APIs and permissions, so I wouldn't be surprised if this is one of those.

glencbz (changes requested):
I only spotted one name that we should change, but everything else looks good :) Great work!

After wayyyy too much time staring at the GCP and Terraform docs, I think I have an understanding of how this is structured. We have the actual resources that traffic flows through on the left, and the way traffic flows between them is defined by the rules on the right:

image

Crucially, there doesn't seem to be a singular thing called a 'load balancer'. Instead, a load balancer is a configuration of routing rules from a frontend to a backend mediated by a Target Proxy (the actual server that terminates the external request).

I'm not sure of the finer details here (e.g. I'm not sure what a Backend Bucket means in terms of actual networking, from what I can tell, it seems to just be a way to use a GCS bucket as a load balancer backend), but hopefully this drawing is useful/matches the understanding we have.

infra/staging/main.tf
glencbz: Actually, I take this back. Naming it `backend_bucket` would be redundant with the 'kind' of thing this is, plus this is for the `frontend-lb` setup, so this is fine.

infra/staging/main.tf
glencbz: Okay, I think I prefer to drop the `_url_map` part (though I won't insist), since:

- If we're viewing the url map, it's redundant. (see also https://github.com/bettersg/call-home/pull/145#discussion_r1139707213)
- If we're viewing the load balancer, it's misleading.

infra/staging/main.tf
```diff
@@ -152,6 +152,99 @@ resource "google_storage_bucket_iam_member" "frontend_bucket_viewer" {
...
+  ]
+}
+
+// Setup SSL certificate to enable HTTPS for load balancer
+resource "google_compute_managed_ssl_certificate" "call_home_frontend_lb_ssl" {
+  name = "call-home-frontend-lb-ssl"
+  project = data.google_project.project.project_id
+
+  managed {
+    domains = ["app2-staging.callhome.sg"]
```
glencbz, line 208: This needs to change between environments, so let's make this a variable.

infra/staging/main.tf
```diff
@@ -203,13 +203,25 @@ resource "google_compute_managed_ssl_certificate" "call_home_frontend_lb_ssl" {
   managed {
     domains = ["app2-staging.callhome.sg"]
   }
+
+  lifecycle {
+    create_before_destroy = true
```
glencbz, line 212: Why do we need `create_before_destroy`?

infra/staging/main.tf
```diff
@@ -203,13 +203,25 @@ resource "google_compute_managed_ssl_certificate" "call_home_frontend_lb_ssl" {
...
   name = "call-home-frontend-lb-proxy"
   project = data.google_project.project.project_id
   url_map = google_compute_url_map.call_home_frontend_lb_url_map.id
+
+  ssl_certificates = [
+    google_compute_managed_ssl_certificate.call_home_frontend_lb_ssl.name
+  ]
+
+  depends_on = [
+    google_compute_managed_ssl_certificate.call_home_frontend_lb_ssl
```
glencbz: Why do we need this explicit `depends_on`? This resource depends on `google_compute_managed_ssl_certificate.call_home_frontend_lb_ssl.name
`, so I'd expect Terraform to already compute this dependency for us.

I see a similar pattern elsewhere. Do take a look at those too.

infra/staging/main.tf
```diff
@@ -152,6 +152,99 @@ resource "google_storage_bucket_iam_member" "frontend_bucket_viewer" {
...
+  ]
+}
+
+// Forwarding rule
+resource "google_compute_global_forwarding_rule" "call_home_frontend_lb_forwarding" {
+  name = "call-home-frontend-lb-forwarding"
+  project = data.google_project.project.project_id
+  ip_protocol = "TCP"
+  load_balancing_scheme = "EXTERNAL"
+  port_range = "443"
```
glencbz, line 238: Ah, as of this version I see that we no longer have a rule for `http://`. As a result, http://app2-staging.callhome.sg/ no longer works.

What we probably want is a typical HTTP -> HTTPS redirect. It doesn't look too hard to set up, hopefully.
https://cloud.google.com/load-balancing/docs/https/setting-up-http-https-redirect
https://stackoverflow.com/questions/64375885/google-cloud-forwarding-rule-http-https-using-terraform

glencbz (changes requested):

infra/staging/main.tf
chaitanyabaranwal: Fixed in https://github.com/bettersg/call-home/pull/145/commits/f2aacfb358f174a88a8554b4996bfde3fa5aeb9f

infra/staging/main.tf
```diff
@@ -39,6 +39,11 @@ variable "frontend_bucket_name" {
   default = "call-home-frontend"
 }
 
+variable "website_domain_name" {
+  type = string
+  default = "app2-staging.callhome.sg"
```
glencbz, line 44: Looks good :)

The name `website_domain_name` threw me off at first, but it's pretty accurate - this is the domain we want for the whole app, not just the frontend.

infra/staging/main.tf
```diff
@@ -152,6 +157,107 @@ resource "google_storage_bucket_iam_member" "frontend_bucket_viewer" {
...
+
+// The following resources create a partial load balancer
+// for redirecting HTTP to HTTPS
+resource "google_compute_url_map" "call_home_lb_http_redirect" {
+  name = "call-home-lb-http-redirect"
+  project = data.google_project.project.project_id
+
+  default_url_redirect {
+    strip_query            = false
+    https_redirect         = true
```
glencbz, line 243: Interesting. So this is a simple HTTP proxy that redirects all requests to HTTPS, preserving the route. As a result, the user would get sent to the HTTPS proxy and all of the real routing can commence.

Not sure if this is elegant or Rube Goldberg-y, but it is simple, so I think this works well for us :) Maybe one day we'll notice a way to do this with one proxy, and maybe that'll save us some money, but we'll worry about that when the time comes.

glencbz (approved):
I've checked that https://app2-staging.callhome.sg/ and http://app2-staging.callhome.sg/ work, and the Terraform rules all make sense to me. Great!

As usual, do prefer squash and merge to remove the review-only commits.

PR #146, opened 2023-03-12
personal-calls: fix FIN typo

glencbz:
The 'serial number' portion was copied from work pass serial numbers and should have been changed.


stevahnes (approved):
Gotcha! Approved.

Issue #138, opened 2023-03-05
[Super app] Bring support detail page up to date with Figma (19 Mar)


stevahnes:
Clean up the support detail page and get UX's approval.

PR #149, opened 2023-03-14
CSS Alignment

stevahnes:
Fixes https://github.com/bettersg/call-home/issues/138

**Landing Page**
image

**Support Detail Page**
image


apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -17,10 +18,25 @@ import './SupportDetailWip.css';
...
 function Header(props: { title: string }) {
   const { title } = props;
+  const navigate = useNavigate();
   return (
     <nav className="support-header">
-      <ArrowBackIosNewIcon />
-      <Typography variant="h4">{title}</Typography>
+      <IconButton
+        aria-label="back"
+        sx={{ color: 'text.primary' }}
```
glencbz, line 26: This key is wrong, so it's no-op (the icon is black). Perhaps you meant `sx={{ color: 'primary.900' }}`?

apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -17,10 +18,25 @@ import './SupportDetailWip.css';
...
+        onClick={() => navigate(-1)}
```
glencbz, line 27: I think we should link to the landing page because this always goes 1 back in history regardless of what the previous page is. e.g. if i go to `google.com` then `http://localhost:3000/support-detail`, this will bring me back to `google.com`.

apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -17,10 +18,25 @@ import './SupportDetailWip.css';
...
+      >
+        <ArrowBackIosNewIcon />
+      </IconButton>
+      <Typography
+        component="div"
```
glencbz, line 32: Why component="div"? `div` is semantically meaningless, whereas leaving it unspecified will at least have _some_ meaning.

apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -17,10 +18,25 @@ import './SupportDetailWip.css';
...
+        variant="h4"
+        sx={{
+          margin: '0px',
```
glencbz, line 35: Prefer `0` to `0px`.

apps/super-app/src/routes/SupportDetailWip.css
```diff
@@ -11,15 +11,15 @@
   border-top-right-radius: 2rem;
   flex-basis: 40rem;
   flex-grow: 1;
-  padding: 1em;
+  padding: 0;
```
glencbz, line 14: The default for padding is 0, so let's delete this line.

glencbz (changes requested):

apps/super-app/src/routes/SupportDetailWip.tsx
stevahnes: Hmmm, I think in general the behavior of the back button should be returning to where the user comes from, instead of going to the Landing Page. If we prefer going back to Landing Page all the time, I think a "Home" button will be better.

apps/super-app/src/routes/SupportDetailWip.tsx
stevahnes: No, it's on purpose. The color of the back chevron should match the primary text color. `primary.900` gives the color below:
image

glencbz: Isn't that the colour we want? Am I misunderstanding something?

PR #155, opened 2023-03-18
Aligned CSS with latest Figma as of 19 March 2023

### LINKED ISSUE #138: [Super app] Bring support detail page up to date with Figma (19 Mar)

stevahnes:
Fixes #138 

### What's done?
- Also implemented the back button logic as per UX Designer's request: return to landing page.
- `SupportDetailPage` was renamed from `SupportDetailWIP`.

### What needs to be done?
- Modify the `Theme.ts` text color.
- Solve the scrolling behavior inconsistency between landing page (header also scrolls) and support detail (header is fixed).
- `SupportDetailPage` icon is not always constrained properly (see the size of IRR logo). Of course this can be easily fixed by ensuring all logos are squared up, but let's see what we can do with CSS.


stevahnes:
@glencbz so sorry I messed up the previous branch commit history, so redoing it as a new branch.

Checked with Inez on the back button behavior and as per my understanding as well, it should return to previous page not the Landing Page all the time.

Also, the color for the back chevron should match that of the header text, which is why the color `sx` is `text.primary`.

glencbz:
> @glencbz so sorry I messed up the previous branch commit history, so redoing it as a new branch.

If you're not using it, please delete the old branch.

> Checked with Inez on the back button behavior and as per my understanding as well, it should return to previous page not the Landing Page all the time.

That's okay, but did you specifically mention my concern with navigating to unrelated websites? I want to make sure she knows about this edge case before committing to a direction.

> Also, the color for the back chevron should match that of the header text, which is why the color sx is text.primary.

My comment was also that they should be the same colour, but this is the state of your PR:

image

The back chevron is clearly black, but the header is blue.

apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -17,10 +18,24 @@ import './SupportDetailWip.css';
...
 function Header(props: { title: string }) {
   const { title } = props;
+  const navigate = useNavigate();
   return (
     <nav className="support-header">
-      <ArrowBackIosNewIcon />
-      <Typography variant="h4">{title}</Typography>
+      <IconButton
+        aria-label="back"
+        sx={{ color: 'text.primary' }}
```
glencbz, line 26: I'm very sure you want `pimary.900`. This rule does nothing, the button is black.

image

See apps/super-app/src/routes/SupportDetailWip.tsx:

```
            <PublicIcon
              sx={{ color: 'primary.900', height: '1.5rem', width: '1.5rem' }}
            />
```

glencbz (changes requested):

apps/super-app/src/routes/SupportDetailWip.tsx
glencbz: Okay, I see the problem, we need to adjust `Theme.ts`.

apps/super-app/src/common/Theme.ts
glencbz, line 1: Notice that on L59 (where we expose the color palette to @mui), we didn't include `text.primary`, that's why `sx` doesn't pick it up.

Let's add `text: colors.text` to fix this.

apps/super-app/src/common/Theme.ts
```diff
@@ -28,20 +28,25 @@ const callHomeTheme = createTheme(
...
         fontSize: 20,
         fontWeight: 700,
         color: '#133844',
         fontFamily: 'Hind',
         lineHeight: '28px',
       },
+      h5: {
+        fontSize: 18,
+        fontWeight: 600,
+        color: '#133844',
```
glencbz, line 40: I didn't realise that these were colour constants in the file. Could we use those values instead? e.g. `colors.text.primary` instead of `'#133844`. Having magic numbers makes their meaning difficult to understand, and this will make it easier to change the theme colours if we want.

glencbz (changes requested):

apps/super-app/src/routes/SupportDetailWip.tsx
stevahnes: @glencbz ah good eye, lemme check idk why!!

stevahnes:
@glencbz fixed! Thanks for checking the chevron color, it looked correct on my screen. Probably my screen isn't very color accurate 😢

apps/super-app/src/common/Theme.ts
```diff
@@ -28,20 +28,25 @@ const callHomeTheme = createTheme(
...
+        fontWeight: 600,
+        color: colors.text.primary,
+        fontFamily: 'Hind',
+        lineHeight: '24px',
+      },
       h6: {
         fontSize: 16,
         fontWeight: 600,
-        color: '#297B94',
+        color: 'colors.primary.800',
```
glencbz, line 47: Nice!

apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -187,7 +209,6 @@ export function SupportDetailWip() {
         blurbFacebookLinks={blurbFacebookLinks}
       ></DescriptionSection>
       <CtaSection
-        ctaBlurb={ctaBlurb}
         ctaButtonText={ctaButtonText}
         ctaLink={ctaLink}
         whatsappLogoSrc={whatsappLogoSrc}
```
stevahnes, line 214: @glencbz can you explain what you want to be done here again?

apps/super-app/src/common/components/ServiceCard.tsx
```diff
@@ -16,7 +16,7 @@ export function ServiceCard(props: ServiceCardAbout) {
     <Card sx={{ display: 'flex', marginBottom: '16px' }}>
       <CardMedia
         component="img"
-        sx={{ width: '5rem', height: '5rem', margin: 'auto 16px' }}
+        sx={{ width: '3.5rem', height: '3.5rem', margin: 'auto 0 auto 16px' }}
```
glencbz, line 19: Hm, maybe at this point it would be clearer to specify `margin-[top|bottom|left|right]` separately, but I don't feel strongly about it.

apps/super-app/src/routes/SupportDetailWip.tsx
```diff
@@ -17,10 +18,24 @@ import './SupportDetailWip.css';
...
   const { title } = props;
+  const navigate = useNavigate();
   return (
     <nav className="support-header">
-      <ArrowBackIosNewIcon />
-      <Typography variant="h4">{title}</Typography>
+      <IconButton
+        aria-label="back"
+        sx={{ color: 'text.primary' }}
+        onClick={() => navigate(-1)}
```
glencbz, line 27: As mentioned in the weekly meeting, I think we can merge this PR with `navigate(-1)` if we leave a TODO to check the edge cases with UXers.

apps/super-app/src/routes/SupportDetailWip.tsx
stevahnes: Left a TODO as discussed.

apps/super-app/src/routes/SupportDetailWip.tsx
stevahnes: Left a TODO as discussed.

apps/super-app/src/common/components/ServiceCard.tsx
stevahnes: Hmmm, feels like it'll be too verbose. Shall we leave it as-is for now?

glencbz: IMO verbosity is less of a problem than understandability, but let's leave it.

apps/super-app/src/routes/SupportDetailPage.tsx
```diff
@@ -76,17 +95,18 @@ function Blurb(props: BlurbProps) {
...
 type DescriptionProps = Pick<
-  Twc2SupportDetail,
+  SupportDetailWithLinks,
   'logo' | 'name' | 'website' | 'blurbIntro' | 'blurbFacebookLinks'
 >;
 
 function DescriptionSection(props: DescriptionProps) {
   const { logo, name, website, blurbIntro, blurbFacebookLinks } = props;
   return (
+    // TODO: support logo is not properly constrained in some cases
```
glencbz, line 105: What do you mean by this, and how will we know when we've fixed it?

apps/super-app/src/routes/SupportDetailPage.tsx
```diff
@@ -116,36 +140,41 @@ function DescriptionSection(props: DescriptionProps) {
...
-  Twc2SupportDetail,
-  'ctaBlurb' | 'ctaButtonText' | 'ctaLink' | 'whatsappLogoSrc'
+  SupportDetailWithLinks,
+  'ctaButtonText' | 'ctaLink' | 'ctaLogo'
 >;
 
 function CtaSection(props: CtaProps) {
-  const { ctaBlurb, ctaButtonText, ctaLink, whatsappLogoSrc } = props;
+  const { ctaButtonText, ctaLink, ctaLogo } = props;
+  // TODO: parameterize ctaLogo img tag, because icon can change based on CTA
```
glencbz, line 149: Is this stale? I see that we're already parameterising on `ctaLogo`.

apps/super-app/src/routes/SupportDetailPage.tsx
```diff
@@ -116,36 +140,41 @@ function DescriptionSection(props: DescriptionProps) {
...
+        <img src={ctaLogo} style={{ padding: '0.25rem' }} />
         {ctaButtonText}
       </PrimaryButton>
     </Box>
   );
 }
 
-export function SupportDetailWip() {
+export function SupportDetailPage() {
+  const { support } = useParams();
```
glencbz, line 171: This looks unrelated from the CSS. Can we remove these changes?

apps/super-app/src/services/DetailServices.ts
```diff
@@ -25,23 +25,21 @@ export function getSupportDetailStrings(): SupportDetailStrings {
...
   // separately.
   blurbIntro: string;
-  // Blurb for CTA section.
-  ctaBlurb: string;
   // Text for CTA button.
   // TODO This won't handle images/icons in the button. Figure that out.
   ctaButtonText: string;
   ctaLink: string;
-  whatsappLogoSrc: string;
+  ctaLogo: string;
```
glencbz, line 38: Maybe `ctaIcon`, since it's more of a decorative icon than a logo?

apps/super-app/src/services/DetailServices.ts
```diff
@@ -56,48 +54,133 @@ export interface FacebookLink {
...
-    ctaButtonText: '+65 6297 7564',
-    ctaLink: 'https://wa.me/6562977564',
-    whatsappLogoSrc: '/images/whatsapp-icon.svg',
+    ctaButtonText: '',
+    ctaLink: '',
+    ctaLogo: '',
   };
+  switch (support) {
+    case 'twc2':
+      supportDetailWithLinks.logo = '/images/twc2-logo.png';
```
glencbz, line 110: I strongly feel that this is change brings us in the wrong direction:

a) We want to replace these hardcoded values with Strapi eventually, but this function doesn't emulate a Strapi calling pattern at all. This takes a single string as a parameter and returns 1 of 3 values we care about, but we plan to use Strapi single types for each page, where a single type has its own url and set of properties.
b) It's replacing a declarative, easy to read style with imperative assignments

glencbz (changes requested):
It looks like you might have accidentally included other changes into this PR. Do send them as a separate PR if this was intentional.

apps/super-app/src/routes/SupportDetailPage.tsx
glencbz, line 1: I think a better direction is to turn this into the TWC2 page. Each support page is so different from one another that we probably can't reuse a single page for all of them anyway.

This was what I meant by `TWC2-only` in the PR description in https://github.com/bettersg/call-home/pull/115, and `move each page's things into their own files` in the commit message of https://github.com/bettersg/call-home/pull/115/commits/37b0d2f4be7e3af5824e4552c261bcc24752c93c

apps/super-app/src/routes/SupportDetailPage.tsx
stevahnes: @glencbz I discovered it when I tried adding the IRR logo into the Details Page. It blew up instead of being constrained to the max-height. Likely coz the logo was tall aspect ratio wise.

We can simply square all our logos to circumvent this actually.

apps/super-app/src/common/Theme.ts
```diff
@@ -59,6 +64,9 @@ const callHomeTheme = createTheme(
       grey: colors.grey,
       error: colors.error,
       warning: colors.yellow,
+      text: {
+        primary: colors.text.primary,
```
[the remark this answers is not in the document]
stevahnes: I don't get this, I tried and it doesn't work. Also added `main` field into the `colors` object but it doesn't work either.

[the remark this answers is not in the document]
glencbz: What did you try specifically? I changed this to

```
      text: colors.text,
```

and it compiles just fine for me. I'd expect it to work for you too, because `colors.text`(what I used) and `{ primary: colors.text.primary }` (what you used in this PR) have exactly the same type.

apps/super-app/src/routes/SupportDetailPage.tsx
```diff
@@ -117,29 +139,33 @@ function DescriptionSection(props: DescriptionProps) {
...
 type CtaProps = Pick<
   Twc2SupportDetail,
-  'ctaBlurb' | 'ctaButtonText' | 'ctaLink' | 'whatsappLogoSrc'
+  'ctaButtonText' | 'ctaLink' | 'ctaIcon'
 >;
 
 function CtaSection(props: CtaProps) {
-  const { ctaBlurb, ctaButtonText, ctaLink, whatsappLogoSrc } = props;
+  const { ctaButtonText, ctaLink, ctaIcon } = props;
+  // TODO: https://github.com/bettersg/call-home/pull/144#discussion_r1141414590
```
glencbz, line 147: If possible, do leave your own interpretation of the conversation so that a) readers don't have to make a round trip to reread the conversation b) readers know exactly what you're referencing instead of reading the whole conversation to figure it out.

In this case, I'm not sure what you're referencing because you've already implemented my suggestion here:

> If we want a generic CTA that can accept any image URL, then sure, parameterising is fine

by accepting the `ctaIcon` prop, so I don't think there's anything left to do.

apps/super-app/src/services/DetailServices.ts
```diff
@@ -6,7 +6,7 @@ import { Support } from './SupportServices';
 
 // Data shared by every Support detail page.
 
-export interface ServiceCardAbout extends Support {
+export interface ServiceCardDetail extends Support {
```
glencbz, line 9: This renaming from `ServiceCardAbout` to `ServiceCardDetail` looks like a leftover from the previous PR iteration. I'd prefer to undo this, but I suppose the name makes more sense, so we can keep it if we want.

apps/super-app/src/routes/index.tsx
```diff
@@ -1,3 +1,3 @@
 export { LandingPage } from './LandingPage';
 export { Options } from './Options';
-export { SupportDetailWip } from './SupportDetailWip';
+export { SupportDetailPage } from './SupportDetailPage';
```
glencbz, line 3: Since we aren't using a generic `SupportDetailPage` any more, perhaps we could omit this change for now and reintroduce it when it makes sense.

apps/super-app/src/routes/SupportDetailPage.tsx
```diff
@@ -13,14 +14,29 @@ import {
...
   const { title } = props;
+  const navigate = useNavigate();
   return (
     <nav className="support-header">
-      <ArrowBackIosNewIcon />
-      <Typography variant="h4">{title}</Typography>
+      <IconButton
+        aria-label="back"
+        sx={{ color: 'text.primary' }}
+        onClick={() => navigate(Path.LandingPage)}
```
glencbz, line 28: Did we change course on the navigation?

apps/super-app/src/routes/SupportDetailPage.tsx
glencbz: That works, though I think there is probably an elegant CSS-based solution for this (e.g. setting the `height` of the containing element). We don't have to discuss this now; we can defer this work until we need it.

glencbz (changes requested):
Thanks! I only have small comments this time, let's clean those up before merging.

Also, it looks like the PR description is out of date. If you could update that to reflect the current reality, that would greatly help me understand this PR :)

apps/super-app/src/routes/SupportDetailPage.tsx
stevahnes: Yes we did. There's a new mock. Let's take the alignment for this one later, but get the logic out first.

apps/super-app/src/common/Theme.ts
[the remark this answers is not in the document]
stevahnes: image

apps/super-app/src/services/DetailServices.ts
stevahnes: Oh no, this one is deliberate as I also feel this name makes more sense :)

apps/super-app/src/common/Theme.ts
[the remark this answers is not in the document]
glencbz: I see, so you had:

```
text: {
  primary: colors.text
}
```

but what I meant was:

```
text: colors.text
```

apps/super-app/src/routes/index.tsx
glencbz: >  If we are going to have individual pages, we can rename them to specific providers later?

Yes. I meant that we would rename them to e.g. `SupportTwc2`.

> I just feel that the "WIP" term is misleading. It's not really "WIP" when you finish one page, no?

Yes, we'll get rid of it when we are done by renaming to `SupportTwc2`. If we are done now, we can make that change.

apps/super-app/src/services/DetailServices.ts
glencbz: Sure. In the future, it would be helpful to call out intentional changes in the PR message or in a comment.

glencbz (approved):
Actually, let's just merge this and clean up the `text:` thing later. It's not big enough to be worth holding up this PR. Sorry for the delay.

As usual, squash and merge would be greatly appreciated.

PR #158, opened 2023-03-19
super-app navbar: add external link to personal calls

glencbz:
This requires refactoring the navbar slightly to support external links.

Also, since this adds yet another service to `services/`, rename the services so that they all follow the same pattern. See the second commit for more details. I'm not sure if naming them `Support` is confusing or not, hopefully @stevahnes can chime in.

Reviewers should check that the navbar renders correctly; the colour of the Call Home link kept changing as I was messing with it. I _think_ it's okay now, but let's be sure.

image


apps/super-app/src/components/navbar/Navbar.tsx
```diff
@@ -40,4 +50,24 @@ function NavIcon(props: { link: string; text: string; icon: React.ReactNode }) {
...
 }
 
+function ExternalNavIcon(props: {
+  link: string;
+  text: string;
+  icon: React.ReactNode;
+}) {
+  return (
+    <div className="navbar-icon-container">
+      <Link
```
chaitanyabaranwal, line 60: I am not sure what's the difference between `Link` and `NavLink` but can't we make it consistent across the three navbar buttons? Does the external Call Home link _need_ to be a `Link`? Also, is that why is needs a `color` attribute but the `NavLink` does not?

chaitanyabaranwal (approved):
LGTM, left one clarifying comment. Also ran locally and the navbar renders properly for me!

apps/super-app/src/components/navbar/Navbar.tsx
glencbz: `NavLink` comes from `react-router`. It works on the client side only by interacting with the browser's (history API](https://developer.mozilla.org/en-US/docs/Web/API/History_API) without actually changing the page the browser is on. This is why we need something other than a `NavLink` to send the user to another host (reasonably, the implementers didn't seem to bother adding additional logic to figure out if we were linking to another host either, so you can get urls like `http://localhost:3000/http://localhost:4000`.)

`Link` comes from `@mui`. It's a styled `<a>` tag. This does the 'normal' kind of navigation where you load a new page. It needs `color` because `@mui` applies some default coloring to `Link` that we want to override. We could have used a plain `<a>`, but this is marginally easier (e.g. it gives us a convenience prop to disable the text underlining of a regular `<a>`)and keeps us in the `@mui` ecosystem.

PR #163, opened 2023-04-02
Add infra for load balancer routing to strapi

halfwhole:
Work in progress!

Having some trouble linking up the load balancer path to the strapi GAE app. Requests are failing with 502 Bad Gateway, and in the logs, the error message on the load balancer is `statusDetails: "failed_to_pick_backend"`. Suspect it's either coming from incorrect routing configs, or strapi not having health checks so the load balancer ignores it as it thinks it's 'unhealthy'

### DIFF
--- a/infra/staging/main.tf
+++ b/infra/staging/main.tf
@@ -193,11 +193,75 @@ resource "google_compute_backend_bucket" "call_home_frontend_lb" {
   }
 }
 
-// URL map for the backend bucket
+resource "google_compute_global_network_endpoint_group" "strapi_external_proxy" {
+  provider = google-beta
+  project = data.google_project.project.project_id
+  name = "strapi-network-endpoint-group"
+  network_endpoint_type = "INTERNET_FQDN_PORT"
+  default_port = 90
+}
+
+# resource "google_compute_global_network_endpoint" "strapi_proxy" {
+#   provider = google-beta
+#   project = data.google_project.project.project_id
+#   global_network_endpoint_group = google_compute_global_network_endpoint_group.strapi_external_proxy.id
+#   fqdn = google_app_engine_application.app.default_hostname
+#   port = google_compute_global_network_endpoint_group.strapi_external_proxy.default_port
+# }
+
+# GOAL: try to map https://app2-staging.callhome.sg/api/* routes to https://call-home-staging.as.r.appspot.com/api/*
+# Latest update: it's failing with 502 and `statusDetails: "failed_to_pick_backend"` as seen in the logs when hitting
+# `https://app2-staging.callhome.sg/api/<whatever>`. Not sure why, perhaps it's because the protocol is wrong (HTTP(S) or TCP),
+# or the health checks are not set up correctly (perhaps on strapi's end too) so LB didn't pick strapi as the backend.
+
+# See StackOverflow answer for 502 failed to pick backend: https://stackoverflow.com/questions/63493125/google-cloud-load-balancer-causing-error-502-failed-to-pick-backend
+# See logs at https://console.cloud.google.com/logs
+# GCP docs on setting up LB infra:
+# - Internet NEG (current approach): https://cloud.google.com/load-balancing/docs/https/setting-up-https-external-backend-internet-neg
+# - Serverless NEG (KIV, might require regional NEGs that are not really compatible with URL maps): https://cloud.google.com/load-balancing/docs/https/setting-up-https-serverless
+# Terraform resources:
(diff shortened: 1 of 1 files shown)


infra/staging/main.tf
```diff
@@ -193,11 +193,75 @@ resource "google_compute_backend_bucket" "call_home_frontend_lb" {
...
+# - Internet NEG (current approach): https://cloud.google.com/load-balancing/docs/https/setting-up-https-external-backend-internet-neg
+# - Serverless NEG (KIV, might require regional NEGs that are not really compatible with URL maps): https://cloud.google.com/load-balancing/docs/https/setting-up-https-serverless
+# Terraform resources:
+# - Global NEG: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_global_network_endpoint_group
+# - Backend service: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_backend_service
+
+resource "google_compute_backend_service" "call_home_strapi" {
+  provider = google-beta
+  project = data.google_project.project.project_id
+  name = "call-home-strapi-backend-service"
```
glencbz, line 229: Here's a kind of janky way to get clues about the "right" Terraform configuration:

- Keep the "wrong" terraform resource
- Create the thing you want in Cloud Console (e.g. click through the "Create Backend" workflow and select App Engine etc etc).
- `terraform import` the resource you just created into the "wrong" terraform resource.
-  Inspect it to see what values are different between the "right" and "wrong" versions.
- ???
- Profit

infra/staging/main.tf
```diff
@@ -193,11 +193,75 @@ resource "google_compute_backend_bucket" "call_home_frontend_lb" {
...
+# or the health checks are not set up correctly (perhaps on strapi's end too) so LB didn't pick strapi as the backend.
+
+# See StackOverflow answer for 502 failed to pick backend: https://stackoverflow.com/questions/63493125/google-cloud-load-balancer-causing-error-502-failed-to-pick-backend
+# See logs at https://console.cloud.google.com/logs
+# GCP docs on setting up LB infra:
+# - Internet NEG (current approach): https://cloud.google.com/load-balancing/docs/https/setting-up-https-external-backend-internet-neg
+# - Serverless NEG (KIV, might require regional NEGs that are not really compatible with URL maps): https://cloud.google.com/load-balancing/docs/https/setting-up-https-serverless
+# Terraform resources:
+# - Global NEG: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_global_network_endpoint_group
+# - Backend service: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_backend_service
```
glencbz, line 224: These look useful :) I'll see if there are clues in these docs.

I suspect that you're already on the right track with Regional NEGs (an example with a different serverless service https://cloud.google.com/load-balancing/docs/https/ext-http-lb-tf-module-examples#with_a_backend).

infra/staging/main.tf
```diff
@@ -193,11 +193,75 @@ resource "google_compute_backend_bucket" "call_home_frontend_lb" {
...
+# GOAL: try to map https://app2-staging.callhome.sg/api/* routes to https://call-home-staging.as.r.appspot.com/api/*
+# Latest update: it's failing with 502 and `statusDetails: "failed_to_pick_backend"` as seen in the logs when hitting
+# `https://app2-staging.callhome.sg/api/<whatever>`. Not sure why, perhaps it's because the protocol is wrong (HTTP(S) or TCP),
+# or the health checks are not set up correctly (perhaps on strapi's end too) so LB didn't pick strapi as the backend.
+
+# See StackOverflow answer for 502 failed to pick backend: https://stackoverflow.com/questions/63493125/google-cloud-load-balancer-causing-error-502-failed-to-pick-backend
+# See logs at https://console.cloud.google.com/logs
+# GCP docs on setting up LB infra:
+# - Internet NEG (current approach): https://cloud.google.com/load-balancing/docs/https/setting-up-https-external-backend-internet-neg
+# - Serverless NEG (KIV, might require regional NEGs that are not really compatible with URL maps): https://cloud.google.com/load-balancing/docs/https/setting-up-https-serverless
```
glencbz, line 221: I haven't read through all of this, but I _think_ https://cloud.google.com/load-balancing/docs/https/setting-up-https-serverless#set_up_regional_routing is suggesting that the NEG is regional, and the URL map would point to a given region. If I'm reading that right, then the two aren't incompatible, they're meant to work hand-in-hand.

PR #168, opened 2023-04-16
super-app: drop landing page

glencbz:
/landing-page is the same thing as /, so we don't need it.


chaitanyabaranwal (approved):

chaitanyabaranwal:
Taking the liberty to merge it since it looks good!

glencbz:
Thanks! Appreciate it. In the future, I'd prefer to merge it myself if it's non urgent (in case I have fixups and whatnot), but no harm no foul :)

PR #172, opened 2023-05-07
Update title and favicon for `super-app`

chaitanyabaranwal:
* The new title is `Call Home`.
* The favicon was copied from the `personal-calls/` directory, so it's the same one as the original Call Home website. I am talking to Inez and will update the favicon in case it should be something different!


chaitanyapotti (approved):
Lgtm

chaitanyabaranwal:
Fixes https://github.com/bettersg/call-home/issues/173

PR #219, opened 2023-10-16
Restructure the components inside `apps/super-app`

chaitanyabaranwal:
The new structure looks a bit saner and is like this:

* All the common components (to be used across multiple places) will reside in `src/common/components`. This includes components like `ServiceCard` and `SupportDetailPage`
* All the top-level single-use components (such as landing page, options, navbar) will reside in `src/components`

This PR also fixes the relevant imports and exports!

This PR is part of generalising the service details inside the super-app, where we will load support details from a JSON and then render the details using similar frontend components.

Partly fixes https://github.com/bettersg/call-home/issues/218


stevahnes:
@chaitanyabaranwal wondering if we should move the `SupportDetailPage` now as the implementation is still very much specific in nature? Perhaps we can move that one after we converted it to a shared component?

chaitanyabaranwal:
@stevahnes Yep, that's the intention :) I have a second PR ready already that moves all the TWC2-specific components outside of `SupportDetailPage` so that it's just a generic component. Will send over a PR for that right after this one is merged!

stevahnes:
@chaitanyabaranwal cool then LGTM!

stevahnes (approved):

PR #220, opened 2023-10-23
Split support detail files into different directories, and make `SupportDetailPage` component generic

chaitanyabaranwal:
This PR makes a few changes:

- Splits the `SupportDetail.ts` file into multiple files
    - `ServiceCardDetail.ts` that stores service card data for the different CSOs
    - `ServiceSupportDetail.ts` that stores support detail page data for the different CSOs
    - `SupportDetail.ts` that contains generic components and interfaces for rendering support detail data

- Also refactors the `SupportDetailPage` component to make it more generic – callers can now pass a function that pulls in data for the CSO that is being currently rendered. Previously, it was defaulting to TWC2 content.

Fixes https://github.com/bettersg/call-home/issues/218.


stevahnes (approved):
Wow, this looks good!

PR #235, opened 2023-12-26
Add Carousel component, and add other partners on main app

chaitanyabaranwal:
**You can review the PR commit-by-commit.**

This PR makes a few changes:

* Adds a Carousel component that is used in PBS's page, taken from Steve's commit in https://github.com/bettersg/call-home/commit/8730fa669b102bbe6b246ae2078541d634a0a525.
* Adds a bunch of other services that were present in the staging website.
* Updates routing to be based on service card details – this means all the service cards will have a route for their support detail page automatically.


apps/super-app/src/services/ServiceSupportDetails.ts
```diff
@@ -86,3 +85,339 @@ const PRO_BONO_SG_DETAIL: Record<LanguageOption, SupportDetail> = {
...
+    website: 'https://www.cde.org.sg/wps/portal/cde/home/',
+    blurbIntro:
+      'The Centre for Domestic Employees (CDE) promote the fair treatment of all domestic employees in Singapore through issues resolution, humanitarian aid, social integration and support. We can support you in up-skilling and emotional resilience',
+    carouselSection: [],
+    blurbFacebookLinks: {
+      title: 'Find us on Facebook',
+      imageSrc: '/images/facebook-icon.svg',
+      links: [
+        {
+          href: 'www.facebook.com/cde.singapore',
```
halfwhole, line 184:
```suggestion
          href: 'https://www.facebook.com/cde.singapore',
```
otherwise, the `href` would be treated as a relative instead of an absolute path!

apps/super-app/src/services/ServiceSupportDetails.ts
```diff
@@ -86,3 +85,339 @@ const PRO_BONO_SG_DETAIL: Record<LanguageOption, SupportDetail> = {
...
+      ],
+    },
+    ctaButtonText: '',
+    ctaLink: '',
+    ctaIcon: '',
+  },
+};
+
+export function getPartnerDetail(
+  partner: string | undefined,
```
halfwhole, line 387: nit, really minor thing: for more type-safety, maybe this could use `ServicePath` instead of `string`?

apps/super-app/src/common/components/support-detail-page/SupportDetailPage.tsx
```diff
@@ -214,6 +214,7 @@ export function SupportDetailPage({
         blurbIntro={blurbIntro}
         blurbFacebookLinks={blurbFacebookLinks}
       ></DescriptionSection>
+      <Carousel carouselSection={carouselSection}></Carousel>
```
halfwhole, line 217: unfortunately the carousel element still renders for partners who have an empty `carouselSection`, which leads to some content on the bottom being blocked out:
Screenshot 2023-12-31 at 10 33 02 PM
perhaps we could try to avoid rendering it unnecessarily by making `carouselSection` nullable, or instead just doing something like this?
```suggestion
      { carouselSection.length > 0 && <Carousel carouselSection={carouselSection} /> }
```

halfwhole (approved):
yeehaw lgtm! maybe steve should verify the final copy and give the green light, but otherwise apart from one issue (empty carousel rendering) it looks and works great. thanks for adding all the partners :)

chaitanyabaranwal:
Thanks for the review @halfwhole! Addressed your feedback, merging this PR now :)

